PR #35, opened 2022-07-05
Feature/add scraped content

kbondarev:
Adds all the scraped data by [show-scraper](https://github.com/JupiterBroadcasting/show-scraper) for the current shows.

Still missing archived shows and [coder-radio 1-54](https://github.com/JupiterBroadcasting/show-scraper/issues/5)


kbondarev:
Oh and I guess these two directories could be removed (nothing references the files there anymore):
- `static/images/guests/`
- `static/images/hosts/`

I left them because I thought maybe those avatars are more preferable for some of the hosts than the one's the scraper pulled from the Fireside websites. If so, just overwrite the images in `static/images/people` (there's two version of each avatar)

PR #38, opened 2022-07-06
Add hover tilt effect to the cards

reesericci:
I added a hover-tilt effect to the card grid on the homepage.﻿

### DIFF
--- a/themes/jb/assets/js/vanilla-tilt.js
+++ b/themes/jb/assets/js/vanilla-tilt.js
@@ -1,0 +1,519 @@
+'use strict';
+
+var classCallCheck = function (instance, Constructor) {
+  if (!(instance instanceof Constructor)) {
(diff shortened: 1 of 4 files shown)


reesericci:
Screencast:
[Screencast from 07-05-2022 09:43:12 PM.webm](https://user-images.githubusercontent.com/19589006/177456879-923222cf-6486-44a0-b214-dc6d2bb64fd3.webm)

kbondarev:
This reminds me of compiz :stuck_out_tongue_closed_eyes:
I think maybe a more subtle effect on mouse hover would be better. For example, slightly changing the background color, or adding shadows around the card to make it look like it's slightly elevated.

kbondarev:
I agree, the less third party libraries the better.

reesericci:
Fair enough

PR #39, opened 2022-07-06
Add workflow to automatically scrape & commit latest shows

kbondarev:
Uses [show-scraper](https://github.com/JupiterBroadcasting/show-scraper) to scraper twice a day: at midnight and noon UTC.

Can also be started manually.


kbondarev:
For some reason this didn't trigger :|

The workflow was suppose to run at midnight UTC (~51 minutes ago) but it didn't... any ideas why?

Issue #106, opened 2022-07-15
Make the media part of the episode pages look better (suggestions)


kbondarev:
Let's put some ideas here about how to improve the design of the episode pages.

---

I'm thinking of using these [tabs](https://bulma.io/documentation/components/tabs/) to split the various podcast formats.

Something like this (not sure about the labels, would be great to hear ideas):

| **Podverse Player** |  **PeerTube** (recorded live stream) | **YouTube**  |  **Direct Download Links** (video/audio files different formats and quality)

The default tab would be opened on the podverse player. PeerTube and Youtube would only be shown on condition - only if the link exist.

kbondarev:
Here's a list of all the components I can think of that ideally will be on the screen:

- Title
- Embedded Players
  - Audio - Podverse (or something else?)
  - Video - YouTube
  - Video - PeerTube (?)
- About episode
- Sponsors
- Membership/Support Show 
- Boost form
- Some metadata info
- Hosts + Guests
- Links to subscribe + Direct downloads
- Tags
- Show links

Maybe change the layout to be something similar/inspired by to Fireside:

![image](https://user-images.githubusercontent.com/7744333/179170881-65302529-8857-40ec-b9cc-06e4001421f3.png)

Would be nice to add this form somewhere (#36):

![image](https://user-images.githubusercontent.com/7744333/179171469-c4bbd80e-7950-4a7a-8329-1895745dc3f7.png)

Lord0fBytes:
Was playing around with this a little bit. Had a concept design for something to put focus on the embedded audioplayer. 
Also re-arranged some text to put it at the top with more importance. 

<img width="1057" alt="Screen Shot 2022-07-16 at 9 22 22 PM" src="https://user-images.githubusercontent.com/7501413/179380025-e88947e9-e40c-4c3b-8e83-e6b502526aef.png">

kbondarev:
@Lord0fBytes: Looks great so far! Much better than before.
 I was about to work on it tomorrow but I guess I’ll leave this to you and focus on other things :))

Regarding the player, take a look at this PR, we are most likely gonna switch to Podverse player:
#105

Lord0fBytes:
Ya I was just trying out one design. I feel like there are a lot of way to go with this since it's pretty big. 
Hoped to spark creativity with it. 

I like the fireside post you put above for inspiration. 

I was making myself familiar with the podverse player for when the PR gets pushed. 

[...]

PR #133, opened 2022-07-21
[WIP] Overhaul to the Episode Page (Issue 106)

Lord0fBytes:
This is the beginning of the new episode page. We are tracking details on #106.
This is a work-in-progress PR so that I can get the new look into production. We would like to break down each section for individual issues to help spread the work. 

@StefanS-O I used `.Scratch` variables in the page. I hope this is OK to use since we call the title a bunch of times and I figured calling the conditional once would be clean. 

Concept Image: 
Concept Ver1 1


elreydetoda:
Also, I love how adaptive this looks on mobile! 🚀

Lord0fBytes:
@elreydetoda Ya there are still a couple things like the breadcrumb isn't perfect, but thats a tough one with the title's having dynamic sizing.

elreydetoda:
Hmm...So, that's an interesting contemplation...I added the episode title to the breadcrumb thing so that way people had a contextual understanding as to where they are in the website when they're on a single show.

But...we could just not have the show name at all, and only just put `Shows / Coder Radio`? I don't know how much that decreases the UX, but might be an option.

Issue #86, opened 2022-07-12
hosts - complete current hosts for each show, some hosts missing


gerbrent:
most shows have an incomplete list of hosts presented on Show page at present.

For example (not exhaustive list):
* LUP - missing Brent
* Coder - missing Wes (was the host in the back catalogue)
* Extras - missing many hosts

Could be best to refer to current JB website/subsites for correct hosts.
Or, better..... can we automate based on hosts scraped? 

input @kiro? @StefanS-O ?

kbondarev:
It seems that for the show page (e.g. [lup](https://jupiterbroadcasting.net/show/linux-unplugged/)) it takes the hosts data from the `_index.md` file
My first instict was to, I agree with @gerbrent  and  automatically aggregate all the hosts that are listed in the episode files.

But then I realized the proper way to look at it is this:
- `hosts` in the _index.md are the **current hosts of the show**. For example Noah was a LUP host long time ago, but maybe it's exactly right to put him the top of the page there as the "host" of LUP
- `hosts` in each episode file are hosts specific in this episode

---

Maybe what we could do is have the title be something like "Current Hosts" and some interactive UI element to show "All current and past hosts". It could be a link/button or maybe 2 tabs. And by default we show the "Current Hosts", but we would also allow the visitor to switch the view to see all the hosts (including past ones)

In the background it would work like this:
- **Current Hosts**: the hosts that are set in the `_index.md` file of each show
- **All Hosts**: automatically aggregate all the hosts from all the episode file of the show

kbondarev:
LUP has "retired hosts" section: https://linuxunplugged.com/hosts

kbondarev:
I mentioned Noah, and looked around a little and realized that he's not in LUP hosts/guests at all. For example, he does host episode but not defined as one: https://linuxunplugged.com/112
We do have him on LAN as a guest: https://linuxactionnews.com/guests/kernellinux

Makes me think that maybe there are a lot of past episode that didn't define hosts/guests properly. Probably a separate issue tho

gerbrent:
all of this is well thought-out and super valid points.

I love your idea of `Current Hosts` and `Past Hosts` - makes alot more sense from a weekly listener standpoint who might now be familiar w the back catalogue or the last decade of JB history.

gerbrent:
Another option - let's call it the `Seperate Page Option`:

* Current hosts are featured top right of the Show page.
[...]

PR #150, opened 2022-07-21
:card_file_box: Fix the active show hosts lists for each show

kbondarev:
Related to #86  (maybe closes?:eyes:)


gerbrent:
so...many.... gitmoji! love it ; )

PR #114, opened 2022-07-18
#88 - Fixed Background Coloring in Navbar hover; Added Shows/People Dropdown options

Lord0fBytes:
Updates the work done on #100 
Resolves #88 

 In PR100 'Guests' was added as a menu item. After discussion it would be helpful to have it in a 'People' sub-menu. 
Added this functionality to this PR. 
Added the shows to a 'Shows' sub-menu

Fixed the hover background coloring in the `components/navbar.sass` file. 

Screen Shot 2022-07-18 at 12 49 35 PM
Screen Shot 2022-07-18 at 12 43 35 PM

Issue #88, opened 2022-07-12
menu - list Guests page in menu


gerbrent:
Currently no way to access a list of guests via navigation.

There is access to https://jupiterbroadcasting.net/guests/ if you know the URL.

Lord0fBytes:
I am working on this issue currently.

Lord0fBytes:
I just realized this was resolved in PR #100 
While working on this I was looking at a slightly different way of doing it. @gerbrent What do you think about the menu having child items?

![Screen Shot 2022-07-14 at 7 04 58 PM](https://user-images.githubusercontent.com/7501413/179115904-cba8c569-afb0-49b0-b3e6-7d11e14829ec.png)

People might not be the right wording, but it was a test. We could do the same for shows as well.

gerbrent:
> I just realized this was resolved in PR https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/100

ahh, sorry I didn't catch that sooner. Still catching up here from quite the week.. (of travels)

I think a sub-menu us certainly an option for likely less-used pages such as "People", as you suggest. I'm hesitant to use it *too* often, however in your example above it's perfectly suited and a nice way to reduce the number of menu items. Nice!

gerbrent:
And agree 100% - would be a great option for the "Shows" menu item. Yes please!

gerbrent:
Also noting the white submenu colour - but I'm sure you noticed that already..... ; )

Lord0fBytes:
> Also noting the white submenu colour - but I'm sure you noticed that already..... ; )

Yup lol. It was just a demo to explain what was in my head 😁

PR #156, opened 2022-07-22
Issue #88 - Fixed Navbar Hover Background Color (white to dark grey)

Lord0fBytes:
Fixed the background color for navbar items when you hover them from white to dark grey. Verified it works on mobile as well. 

This resolves the request from #114


gerbrent:
woot! thanks for getting to this so quickly!

All I can say is..... 🔥 !!

Issue #13, opened 2022-06-08
Matrix info update - change instructions to Matrix.org server accounts


gerbrent:
https://linuxunplugged.com/matrixinfo currently lists signon instructions for accounts under the JB Matrix server, which are now disabled. Should migrate the instructions toward Matrix.org signups.

gerbrent:
I've been lagging on this one, as mentioned in [LUP469](https://jupiter.tube/w/n9oSKDs6TVYrs8kv8xeZqx) (to be releases) and listeners are suffering for it.

Would someone else be willing to:

* [ ] update the Matrix info found at https://linuxunplugged.com/matrixinfo and populate ./community/matrix
* [ ] then JB can use that updated guide to update https://linuxunplugged.com/matrixinfo in turn
* [ ] https://linuxunplugged.com/matrixinfo will become redundant and thus potentially removed or redirect to here on JB.com 1.0 publish date.

PR #196, opened 2022-07-29
Add detailed Matrix Info

xPMo:
I have some screenshots, in case a more detailed Element walk-through is desired. The matrix.to links have decent onboarding for new users, so I don't think it's necessary anymore, especially since we don't have open registration.

Ref: #13


gerbrent (approved):
love how you structured this, really nice. Thanks!

Issue #162, opened 2022-07-22
All Shows feeds - include in hugo site


gerbrent:
We need a place to link to the All Shows feeds 

* [All Shows Feed - Audio](http://feeds2.feedburner.com/JupiterBroadcasting)
* [All Shows Feed - Video](http://feeds2.feedburner.com/AllJupiterVideos)

related to #66

aaronrea:
At first pass, the most reasonable spot for these links seems to be at the top of the `Shows` header menu, imho. Any thoughts?

elreydetoda:
If we do that, I think we should just treat it like a show then.

i.e. add it to the carousel, it has it's own page, and populate it with all the episodes (which could just be references to the other pages on the site (which'll be easier once we finish #110 (trying to finish that this week, but work has been a lot)))

One thing I'm actually curious about is, does the JB all shows feed have all episodes from forever or is it only active shows (not archived)?

CGBassPlayer:
The all shows feed does seem to have archived shows in it.
![image](https://user-images.githubusercontent.com/34293941/181749593-e8bb3d78-57d4-4d8e-b3d3-8649c0bac795.png)

gerbrent:
Oh yeah, that All Shows feed has been around *forever*.

I wonder if it's worth pondering a little strategy there..........

gerbrent:
> If we do that, I think we should just treat it like a show then.
> 
> i.e. add it to the carousel, it has it's own page, and populate it with all the episodes (which could just be references to the other pages on the site (which'll be easier once we finish #110 (trying to finish that this week, but work has been a lot)))

Not a bad idea at all. Hmm! Worth pondering... thoughts anyone on why this might be troublesome or otherwise have unintended considerations?

elreydetoda:
> troublesome or otherwise have unintended considerations?

Only thing that I could think of, if we do it exactly the way I described, where it lists out all the shows but when clicked it goes to the actual link for the show. 


i.e. the breadcrumb would say Shows > Self-Hosted > (self-hosed-episode-title)

So it would break the expected behavior (based on all the other shows) that it leaves a breadcrumb back to the all shows feed. Which honestly I don't think would be that bad because they could just hit the back arrow, but I'm not a UX person 😅

reesericci:
All Shows isn't a show, and shouldn't be treated as such from a UX perspective.

elreydetoda:
[...]

PR #236, opened 2022-08-06
Clean up latest episodes on homepage

reesericci:
I removed pagination from Latest Episodes on the homepage, as it shouldn't be an ever-expanding list of episodes - that belongs to an "All Shows" page. Following in line with that, I shortened the amount of episodes shown on the homepage from 15 to 6.

Screenshot:

image


graemearthur:
Think this would work with a more episodes link at the end for those who are looking for all episodes

gerbrent:
Well, you see, this was specifically a feature request I made several weeks ago when the home page was instead the entire JB catalogue - to show many recent episodes, but not all, and w natural pagination.

#### My perspective and thus motivation for that ask is this:

* a new-to-JB listener will likely arrive at the first page, and be in the mood to explore. That can be done either by browsing the menu and sub-menus, or by scrolling. On mobile, scrolling down a page or two is a very intuitive action, and you thus get a good sense of JB's recent offering with more than 6 recent episodes on offer. 
* A seasoned JB listener will know there are several shows, and can either dig into the menu for the specific show they're after, or more often than not browse the last few episodes on offer. 
* If you are a casual JB listener (in contrast to a listen-to-everything type listener) and thus haven't listened in some time, exploring recent topics across all shows is a great way to re-engage said listener.
* with such a small recent-episodes offering on the front page, I get a small sense that perhaps this subset is the only offering on hand. It confuses the "Shows Carousel" with the "Recent Episodes" view, as they have the same artwork, similar shapes, similar number of items. The distinction between both feels obscured at that point in my opinion.

#### scenarios in which 6 or less is not ideal:

With only 6 recent episodes shown, there are regular sub-optimal scenarios where a show is not present on the home page, despite having a relatively recent episode:
* LAN251, Coder476, LUP468, Office8, LAN250, Coder475 - Where's the latest SSH?
* Office 9, LAN252, Coder477, LUP469, SSH76, LAN251 - Only one LUP present, our biggest show?

#### Alternatives
I would welcome discussion and perspective around alternatives - a `more` button might just be the right touch - however I do not believe the above commits and sample without pagination or `more` button is the answer.

Another consideration: A `more` button would thus require an "All Shows" page, which is discussed and strategy not yet decided in #162

gerbrent:
In conclusion, I do not feel this should be merged until more discussion can happen and a decision be made on approach.

reesericci:
Could we increase the count of episodes from 6 to 9?

—reese

On August 18, 2022, Bananaman ***@***.***> wrote:
> In my opinion this should not be merged for the reasons you stated
> above @gerbrent <https://github.com/gerbrent>. Tbh I'd close this PR
> too as noone else seems to have strong opinions in the opposite
> direction.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/236#issuecomment-
1219419301>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AEVOPDVTTVRRDAZKN6XVOI3VZYSOJANCNFSM55ZGYP6A>.
You are receiving this because you authored the thread.Message ID:
***@***.***>

gerbrent:
@reesericci: @ChrisLAS was interested in the same, though i think we should keep pagination...

reesericci:
I think pagination _is_ hurting the design because "Latest Episodes" shouldn't be an ever-flowing stream of content & makes the design less elegant imo.

ironicbadger:
Your opinion is noted. Thank you!

reesericci:
Is that supposed to be passive-agressive?

—reese

On August 18, 2022, Bananaman ***@***.***> wrote:
> Your opinion is noted. Thank you!
>

—
Reply to this email directly, view it on GitHub
<https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/236#issuecomment-
1219811102>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AEVOPDSQ6X3VCXK4MIUOG4DVZZ7EVANCNFSM55ZGYP6A>.
You are receiving this because you were mentioned.Message ID:
***@***.***>

gerbrent:
> I think pagination _is_ hurting the design because "Latest Episodes" shouldn't be an ever-flowing stream of content & makes the design less elegant imo.

I feel like a "more" button that links to a (currently non-existent) `All Episodes, sorted by latest` page would be a really happy solution to this (and potentially to partially #162 aswell). Though it can be said that it's a small detail for Milestone 1.0, and our collective efforts are likely better spent elsewhere on more pressing and essential tasks.

gerbrent:
https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/236#issuecomment-1207317664

see here also as a reminder of all my particular reasoning behind the above.

ironicbadger:
> Is that supposed to be passive-agressive? —reese

Not at all :) Just wanted to let you know that you've been heard even if we don't go with your suggestion.

reesericci:
Just changed from 15 to 9 and resolved merge conflicts!

themes/jb/layouts/index.html
```diff
@@ -62,9 +63,6 @@ <h1 class="title">Latest Episodes</h1>
 
       </div>
     </div>
-    <div class="container">
-          {{ partial "helper/pagination.html" . }}
-    </div>
```
gerbrent, line 66: We would like to keep the pagination. Could you keep this section please?

gerbrent:
```suggestion
    <div class="container">
          {{ partial "helper/pagination.html" . }}
    </div>
```

gerbrent (changes requested):

reesericci:
Ok just implement your changes!

gerbrent (approved):

gerbrent (approved):
Tested and all is perfect! Thanks @reesericci !

Issue #233, opened 2022-08-06
Live page - why a separate config/site?


gerbrent:
The menu, footer at https://jb-live.jupiterbroadcasting.net/ differ from the main Hugo site.

This leads me to question:
* Why is the Live page necessarily a separate hugo site?
   * seems this would only add to inconsistencies..

i.e. we have a dedicated [config.jb-live.toml](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/config.jblive.toml) - is that really necessary?

#### Expected
The Live page should feel cohesive w the rest of the site, asthough it's just another page on the JB main site..

gerbrent:
added to #233

reclaimingmytime:
I agree, I would also prefer the live page to be part of the main page.

When switching between both sites, the change of the menu items in the navigation becomes confusing and you need to find your orientation again on the new page.

elreydetoda:
Don't have time to work on this yet, but this'll probably help with the consistency: https://gohugo.io/hugo-modules/configuration/#module-config-mounts

In hugo you can "mount" different types of pages to other locations (think of it like a symlink), so you should be able to do that for different files/folders.

gerbrent:
oh sorry, my description wasn't clear enough. Adding to the initial comment to clarify.

gerbrent:
> i.e. we have a dedicated [config.jb-live.toml](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/config.jblive.toml) - is that really necessary?

^ added this

gerbrent:
> @gerbrent, so are you saying (with this issue) that you'd like the url to be something like: new.jb.com/live ?

this url would be completely acceptable, but is not the main intention of this ticket.
The main intention is to address the many little inconsistencies and errors we are getting from having `jb-live` have it's own `config.toml`

PR #248, opened 2022-08-08
removes unneeded actions and comments

ironicbadger:
Small changes but removes old jobs that aren't needed anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,17 +1,12 @@
-# This is a basic workflow to help you get started with Actions
-
 name: CI
 
-# Controls when the action will run.
 on:
-  # Triggers the workflow on push or pull request events but only for the main branch
   push:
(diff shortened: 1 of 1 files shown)


gerbrent:
Quick note to add: This also removed the Coder Radio sub-site build, which... feels like a relic from the MVP days when we assumed we needed a sub-site for each JB Show. I feel the current website features fulfill all the basic requirements for each Show, and a sub-site for each Show is superfluous (and unfortunately our current and historical approach..).

gerbrent:
The same can be said for the JB-Live subsite.... see #233 for suggestion/question related to integrating JB Live into the main site aswell.

StefanS-O:
Yeah it was more a proof of concept, that you can also build subsites from the same codebase with the same theme that get's extended.

StefanS-O (approved):

PR #249, opened 2022-08-08
Infra/build improvements

ironicbadger:
Removed all the logic from the makefile pertaining to logic binaries of hugo. The production site will be deployed as a container so it makes more sense to build and develop the site in the same environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,3 @@
-FROM registry.gitlab.com/pages/hugo/hugo_extended:latest
+FROM klakegg/hugo:0.101.0-ext-alpine
 CMD [ "serve", "-D", "--bind", "0.0.0.0" ]
 ENTRYPOINT [ "hugo" ]
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,3 @@
-SHELL := bash
-DOCKER_CHECK := $(shell command -v docker > /dev/null 2>&1 ; echo $$? )
-ifeq ($(DOCKER_CHECK), 0)
-BIN := docker
-else
-BIN := hugo
-endif
-
-ifeq ($(BIN), hugo)
-BIN_FLAGS := serve -D
-else
-BIN_FLAGS := container run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest
-endif
-
-dev: deps
-	$(BIN) $(BIN_FLAGS)
-
-deps:
-	if [ "$(BIN)" == hugo ] ; then \
-		hugo version | grep 'extended' >/dev/null || (echo "You don't have the extended version of hugo, please install it" && exit 1); \
-	else \
-		docker build --rm -f Dockerfile -t jb_hugo:latest . ;\
-	fi
+dev:
+	docker build --rm -f Dockerfile -t jb_hugo:latest .
+	docker run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest
--- a/README.md
+++ b/README.md
@@ -24,41 +24,15 @@ Built with Hugo and deployed with Github Actions
 
 Demo: https://jupiterbroadcasting.net
 
-### Setup
+### Development Environment Setup
 
-#### Using Hugo binary
-You must install  the Hugo Extended Sass/SCSS version.
+The site is deployed to production as a container and therefore we recommend development is done using docker. A convenient command in the makefile is provided to help you get started. If you're new to docker [start here](https://docs.docker.com/get-docker/).
 
-Install Hugo: https://gohugo.io/getting-started/installing/
+To build and run the site run
 
-Hugo Extended Linux binary: https://github.com/gohugoio/hugo/releases
+    make dev
 
-Start the development Server (rebuilds on every filesystem change)
-
-`hugo server -D`
-
-#### Using Docker
-
-To build and run the docker image:
-`make dev`
-
-#### run for different Site
-
-`hugo server -D --config config.coderradio.toml`
-
-to clean the module config
-
-`hugo mod clean --all`
-
-build
-
-`hugo -D --config config.coderradio.toml`
-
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM registry.gitlab.com/pages/hugo/hugo_extended:latest
+FROM klakegg/hugo:0.101.0-ext-alpine
```
elreydetoda, line 1: So, it looks like this is just a random individual's docker container.

While it looks like it was pinned in [Hugo's forum](https://discourse.gohugo.io/t/klakeggs-wonderful-hugo-docker-image-update-thread/20347), is there any reason why you went with that instead of Gitlab's?

Just looking at the forum they don't appear to be a Hugo maintainer and don't have a mod icon next to their name.

elreydetoda:
This'll be nice to simplify logic for local development 😅

I am curious though, should we maybe mention something about "...while discouraged, you can still use the hugo binary locally. We won't be able to help troubleshoot if used outside of the normal development workflow"?

Because I think one of the nice things about hugo is that it's just one go binary away, but if you're going to consistently try and do development for the project you'll probably choose the tested route of docker (since it'll be prescribed 8n the docs).

Dockerfile
ironicbadger: The image is the one listed on the Hugo project website, that's why. In lieu of an official image it seems prudent to use the closest thing we have to one.

This won't be the final production image I don't think. Only dev / CI. 

So until there's a solid technical reason for one image over the other my view would be to use the only image Hugo themselves promote on their website.

ironicbadger:
> I am curious though, should we maybe mention something about "...while discouraged, you can still use the hugo binary locally. We won't be able to help troubleshoot if used outside of the normal development workflow"?

We could. But I think anyone who cares enough about that will already know. Also, the simpler the docs the better. Less is more.

Dockerfile
elreydetoda: > So until there's a solid technical reason for one image over the other my view would be to use the only image Hugo themselves promote on their website.

So, my only qualm with using the other image is from a security perspective, that even though they are promoting it. It's not backed by one of their team member or by something/someone who's got more of a reputation.

One day their project could be perfectly fine, and then the next they could pull a [protestware stunt](https://arstechnica.com/information-technology/2022/03/sabotage-code-added-to-popular-npm-package-wiped-files-in-russia-and-belarus/) the next day and if JB relies on that for their main website deployments it could modify/alter the deployment.

While with the Gitlab image, I'm sure technically it *might* be possible, but there is a significantly lower chance of things happening to it because it's produced by Gitlab employees on Gitlab infrastructure.

Also, the hugo docs do recommend using that image, but it's just on the Gitlab deployment page: https://gohugo.io/hosting-and-deployment/hosting-on-gitlab/

elreydetoda:
Ya, makes sense, just figured I'd throw it out there for consideration 🙂

Dockerfile
ironicbadger: The image itself doesn't really matter too much I guess.

From my time at Linuxserver.io I gravitate towards Alpine images. I'll probably be around to fix any image issues that crop up too so I'm not too worried about that aspect. Though I would prefer Hugo to just make an image overall.

elreydetoda: > Though I would prefer Hugo to just make an image overall.

Ya, I'd definitely prefer this! :sweat_smile: 

Also, the gitlab image is an alpine image. I thought it was, but I couldn't confirm till I got to my desktop (on mobile before).
image

elreydetoda: So, based on all that we've talked about I'd recommend just keeping the gitlab image:

```suggestion
FROM registry.gitlab.com/pages/hugo/hugo_extended:latest

```

elreydetoda, line 1:
```suggestion
FROM registry.gitlab.com/pages/hugo/hugo_extended:latest
```

elreydetoda (changes requested):
So, based on all that we've talked about I'd recommend just keeping the gitlab image, instead of the other image.

Dockerfile
elreydetoda, line 1:
```suggestion
FROM registry.gitlab.com/pages/hugo/hugo_extended:latest
```

elreydetoda (changes requested):
So, based on all that we've talked about I'd recommend just keeping the gitlab image.

Dockerfile
ironicbadger: Is there a poll feature here? Because I still stand by my original argument that Klackett is the image the Hugo project themselves recommended and comes from Docker Hub (maybe more reliable than gitlab registry? I don't know).

Perhaps we can just use reactions. 

- [ ] Thumbs up for Hugo project recommendation (Klackett)
- [ ] Smiley face for gitlab image

elreydetoda: Sounds good, we can try and do the emoji reactions for voting. Did you see my comment that they actually do recommend the gitlab one as well?

> Also, the hugo docs do recommend using that image, but it's just on the Gitlab deployment page: https://gohugo.io/hosting-and-deployment/hosting-on-gitlab/

---

> maybe more reliable than gitlab registry?

Ya, I don't know about this one, I'd imagine Gitlab's registry would have to be pretty rock solid, since they tout themselves as the "The One DevOps Platform" for businesses. I'd argue they'd probably have to be more resilient, because of how many businesses they probably support compared to docker is still trying to find a way to become profitable. Plus the docker registry has their [rate limiting now](https://container-registry.com/posts/overcome-docker-hub-rate-limit/).

gerbrent:
Great conversations here. I've voted w my opinion, but curious what @StefanS-O thinks overall..

ironicbadger:
I have reverted the image to `registry.gitlab.com/pages/hugo/hugo_extended:latest`.

Dockerfile
ironicbadger: I have reverted the image to `registry.gitlab.com/pages/hugo/hugo_extended:latest` based on the voting results (1:4) in favor.

gerbrent:
🎉 !

Issue #145, opened 2022-07-21
Hosts/Guests pages - make layout identical


gerbrent:
* [ ] Apply changes made previously to Hosts page design/layout apply to Guest pages

For example, see differences:
* [hosts/alex](https://jupiterbroadcasting.net/hosts/alex/)
* [guests/alex](https://jupiterbroadcasting.net/guests/alex/)

### hosts/alex
![image](https://user-images.githubusercontent.com/7018928/180285054-c4dfe85f-467d-4195-a57c-476b31769985.png)

### guests/alex
![image](https://user-images.githubusercontent.com/7018928/180285142-d57dd43e-f8d0-4dc5-b205-c34e51ca5723.png)

gerbrent:
@elreydetoda , @[Dedpy](https://github.com/Dedpy) ... interested in working on this one?

elreydetoda:
Sure @gerbrent , I can take this one.

Lord0fBytes:
@elreydetoda If you are still working on this one, I updated the hosts page in PR #228 to adjust the layout and aspect ratio of the cards. Feel free to copy the layout for the guests or let me know if you want some help.

elreydetoda:
That sounds good, thanks @Lord0fBytes , I'll be sure to checkout the hosts page and match them. 😁

gerbrent:
In the Hugo spirit of "everything only exists once" - is there a different way then to define the Host and Guest pages to always be identical in functionality?

elreydetoda:
Ideally...yes, but I'll need to check in the code (on mobile now) to see what's currently prohibiting that from happening now.

The only thing that I could think of is if the keys in the json are different. Even so, we should be able to handle that just not as easily if they were the same.

I'll work towards that with this issue though.

elreydetoda:
Alright, for now I think I've replicated what @Lord0fBytes did for the host page to the guest page. I'll wait till he replies to #241 to actually make the PR ready to merge, so the pages are consistent (with or without the `-`). Here is a quick screenshot of pages.

Host page
![image](https://user-images.githubusercontent.com/10230166/183336458-987026d8-ef12-488e-a248-14edb0faf2fb.png)

Guest page
![image](https://user-images.githubusercontent.com/10230166/183336472-c97b6a67-e54f-4383-ac33-7ba8863e204b.png)


[...]

PR #252, opened 2022-08-09
Feat/redo json to pages/110

### LINKED ISSUE #110: Redo the `data/` folder to pages
So, I meant to do this before with @StefanS-O's MVP repo, but didn't get to it before everything migrated (sorry :sweat:, life happened ([reference](https://matrix.to/#/!rDlEtonuSythLzmxQL:jupiterbroadcasting.com/$8noW3K5gYRKkcVykVswo8IDv5umzqFQw9z76rjqjWYA?via=matrix.org&via=jupiterbroadcasting.com&via=0x6b.dev))). Also, sorry for it being such a long issue, I just wanted to make it as comprehensive as possible. (Text only (at least for me) is hard to be as clear as possible without being overly verbose, for clarity sake)

### tl;dr:
the current way we're writing "people pages" (using raw json files to propagate everything), isn't as hugo _native_ as it could be. By having things more hugo native, it lends itself to doing things more efficiently and easily modifiable/usable with hugo native functions.

#### Current solution (from my understanding)

Repo is currently using raw json files for all people (hosts & guests).

#### Proposed solution

Instead of using machine readable (granted since it's not that long/large it's easily human readable for technical people) json, using Hugo's native [front matter](https://gohugo.io/content-management/front-matter/) + markdown under the `content/` folder is more what hugo expects (and I'm sure has more optimizations for building).

#### More info

The reason why (IMO) it's important to convert these from json to yaml, is because hugo has builtin functions to help handle pages more natively than it does with raw json.

For example, the [`where` function](https://gohugo.io/functions/where/), allows for easily being able to filter down arrays of Page objects. You can even get really complex with nesting `where` functions, and then you probably don't even have to use an if statement later. That's in comparison to having to load all the json files, and then looping over them all and having them each hit an if statement to see if they match a specific parameter.
This is an example of being able to filter the page objects (with `where`) before you actually start iterating over them: https://github.com/GWC-UNCC/Girls-Who-Code-at-UNCC/blob/6555ad7701ad0e281e2142d4ba518f72cd345b4a/layouts/partials/portfolio.html#L12

In the above example I was able to:
1. grab the portfolio pages (so we could highlight some of them on the main (home) page)
2. Then check if there are any pages whose date is newer than now (upcoming events)
3. Afterwards I was able to grab the first 3, so we could highlight only a few of them.

This allowed me to then easily iterate over them, like a normal list page would do but just inline on the homepage.

<br />

Another great example of using pages instead of json objects is whenever you want to work with pagination (i.e. what I was recommending on #92, instead of filtering (since pagination is native to hugo pages)).  If we use json objects, then we'll have to implement pagination by ourself. While on the other hand if they were pages pagination is builtin to hugo, and hugo has a native paginator (that you can override the colors for themeing).
For example, in a previous website I worked on, since all the posts were pages it was easy to implement pagination with about 3 lines of code:
1. Use this to indicate the pages you're ranging over need to be paginated: `{{ range (.Paginator 9).Pages }}`
2. Reference the content you want (like a normal page reference) and then close the range.
3. include this code block for the pagination functionality: `{{ template "_internal/pagination.html" . }}`

This will automatically handle slicing up all the page that need to be paginated, and placing them on the appropriate pages.
Here is the whole code block I used in my previous project:
https://github.com/Climate-Refugee-Stories/crs-website/blob/50d6f3f2ffcc6ac8d388063e7e159c14c477dbce/layouts/stories/list.html#L11-L17
Here is the live site's functionality: https://www.climaterefugeestories.com/stories/

<br />

Similarly to the last one, when doing list pages you'll get a bunch of native goodies with hugo and using page objects. One such example is that you can order pages with a special front matter field called `weight`. Hugo will look for this field when trying to list the individual pages and will automatically sort the pages based on that weight. It's sort of weird how they implement it, but 0/none = sorting by the next special frontmatter variable (A-Z or date) and then 1 > 2 > 3... So, everything after 1 is decreasing in priority, which makes sense for the average person, but programmers normally think of 0 first (or at least I do :sweat_smile: ).
Another example, all the people are listed under one folder here: https://github.com/Climate-Refugee-Stories/crs-website/tree/50d6f3f2ffcc6ac8d388063e7e159c14c477dbce/content/about.

Normally this was sorting by their name in alphabetical order, but we wanted to list the organizer (Tina) first. So, all we had to do was add the [weight of 1](https://github.com/Climate-Refugee-Stories/crs-website/blob/50d6f3f2ffcc6ac8d388063e7e159c14c477dbce/content/about/Tina-Shull.md?plain=1#L3) to her page, and she automatically is listed first now: https://www.climaterefugeestories.com/about/#our-team

<br />

Lastly, whenever there is a chance for implementing the CMS. If you're using pages for pulling people, at least with [NetlifyCMS](https://www.netlifycms.org), you can reference those pages and their front matter with relative ease via the [relation widget](https://www.netlifycms.org/docs/widgets/#relation).
That widget allows you to propagate information for the user to select from (i.e. hosts or guests when creating a new episode for a show). Here is a picture (using the netlify CMS docs as an example, since they've implemented it) of what it'd look like:

image

Here's how they did it:
1. this is their config for NetlifyCMS, and their intro config yaml: https://github.com/netlify/netlify-cms/blob/0c5efd19d6aa28e1e06a90c221ea0e8ac834ef33/website/static/admin/config.yml#L123-L128
2. They're using YAML [aliases & anchors](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd) to keep their YAML [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself): https://github.com/netlify/netlify-cms/blob/0c5efd19d6aa28e1e06a90c221ea0e8ac834ef33/website/static/admin/config.yml#L16-L33
3. Inside that YAML anchor this is specifically the relation widget: https://github.com/netlify/netlify-cms/blob/0c5efd19d6aa28e1e06a90c221ea0e8ac834ef33/website/static/admin/config.yml#L23-L30

Which lets them propogate the dropdown based on their search terms, to select with the section name that the content should be placed under with their documentation (I was in the [overview section](https://www.netlifycms.org/docs/intro/)).

#### Technical implementation of scraping

Converting the scraper to use the new format shouldn't be that bad to implement either.  Since the yaml library already can handle reading with the `yaml.load_all()` function (according to [this](https://stackoverflow.com/a/34727830)), and the dumping function (similar to dumps for json in python) can write the same way: https://pyyaml.org/wiki/PyYAMLDocumentation#dumping-yaml (1st yaml doc = front matter & 2nd = MD content).

elreydetoda:
Closes #110 , will definitely need people to test this locally first.

Way to test locally:
- pull down my scraper repo and checkout this branch: https://github.com/elreydetoda/show-scraper/tree/feat/json_to_hugo_pages
- point data dir to be the base directory for this repo (change the first part of the volumes in the docker-compose file to be the root of this directory)
  - I'd also recommend you add this to your `.gitignore`, so you IDE doesn't cry when 100s of files get added to your disk
     ```
     content/people
     content/sponsors
     content/show
     tmp/
     ```
- Run the scrapper
- Launch Hugo
- Delete the data folders
- Re-launch hugo (to ensure no residual settings are configured)

We'll also need to coordinate with @kbondarev, so that the scaper can be merged at the same time as the main repository. This'll be important, so that way neither version of the blog is broken (the update script or just the regular blog).

Also, @gerbrent you'll be happy that I was able to unify a bunch of the views (like you mentioned here: https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/145#issuecomment-1207232805) for different things. So, now there is a `partials/people` folder and all the stuff from `partials/hosts/` & `partials/guests/` were merged into that :grin:


kbondarev:
@elreydetoda Great work! I'll try to go over it and test it later on tonight (EST).

kbondarev:
@elreydetoda It all looks great to me, except one thing: the names of the ppl aren't showing. I think it might be related to your last commit. Maybe the `name` in people should also be renamed to `title`. I tried that for some but it didn't help. You probably would know more what's going on..

elreydetoda:
LOL... 🤦

I forgot to push my last commit that I made for that into the scraper repo :sweat_smile: 

I just pushed it now :disappointed: https://github.com/JupiterBroadcasting/show-scraper/pull/19/commits/2bd0536d519cd3206b6ec770e710fa75ed9a0c17

Nice catch @kbondarev :rocket:

kbondarev:
Also there's a page available under `/people`. Any way to suppress it? I feel like google might index it.

Or should we need to put something there? like a list of all the people?

elreydetoda:
Ya, I could whip up a quick list page that would just list all the people that have ever been on the JB network. I think that'd be pretty cool actually :slightly_smiling_face:

kbondarev:
Maybe then add some kind of an indicator if they're a host, or split the page into two section `Hosts` and `Guests`

elreydetoda:
Ya, I just realized that the People menu option appears clickable but isn't configured (so links to the current page you're on), so I think it'd be perfect to have that linked to the people page. :sweat_smile:

elreydetoda:
Alright, fixed the merge conflict (after rebasing from main) and tomorrow I'll finalize this PR and set it to "Ready for review". Last things (chatted with @kbondarev for troubleshooting something) are:

- [ ] add the actual people files & sponsors
- [ ] delete that data folder

> Ya, I could whip up a quick list page that would just list all the people that have ever been on the JB network.

:arrow_up: - is [done](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/252/commits/27711a3416a9c2cab50db0314a8ea287c19f217f) BTW

gerbrent:
nice work you two!

kbondarev:
Great Job @elreydetoda :rocket:  :rocket: :rocket: 

This was merged via #261 

I'll make a separate issue for the pagination improvement.

elreydetoda:
Thanks for pick up the rest of this issue, sorry work has been pretty crazy and I was planning to try and work on it this morning but @kbondarev beat me to it :sweat_smile: :grin: 
Thank you again for the assist @kbondarev for getting this over the finish line. :partying_face: :confetti_ball: :tada:

PR #253, opened 2022-08-09
Add irc details and convert link to button

### LINKED ISSUE #210: IRC in menu, Community page, footer - leads to a bland page
the "IRC" menu item AND card on the [Community page](https://jupiterbroadcasting.net/community/) AND footer item leads to an irc-dedicated page at https://jupiterbroadcasting.net/community/irc/

This page is rather bland.

#### We should either:

A. link directly to [Kiwi IRC](https://kiwiirc.com/client/irc.geekshed.net/?nick=JBLive_?&theme=cli#jupiterbroadcasting)

or...

B. link to the Live! page (#163 ) with the IRC embed.

or...

C. Embed the Kiwi IRC into the ./community/irc page

#### todo:
* [ ] change IRC sub-menu item behaviour
* [ ] change IRC card on [Community page](https://jupiterbroadcasting.net/community/) behaviour
* [ ] change IRC footer link behaviour

thoughts? feedback? preferences?

reclaimingmytime:
Screenshot 2022-08-09 at 16 52 11

Closes #210

### DIFF
--- a/content/community/irc.md
+++ b/content/community/irc.md
@@ -10,4 +10,14 @@ comms_type = "network"
 direct_link = ""
 +++
 
-https://kiwiirc.com/client/irc.geekshed.net/?nick=JBLive_?&theme=cli#jupiterbroadcasting
+<a href="https://kiwiirc.com/nextclient/irc.geekshed.net/?nick=JBLive_?#jupiterbroadcasting" class="button is-primary">Launch Web Client</a>
+
+<div class="card mt-5">
+    <div class="card-content">
+        <div class="content">
+            <p class="card-header-title">Prefer to use your own IRC client?</p>
+            Server: <strong>irc.geekshed.net</strong><br>
+            Channel: <strong>#jupiterbroadcasting</strong>
+        </div>
+    </div>
+</div>


gerbrent:
Related but slightly tangential question:

We will now have IRC information listed in two places:
* [./community/irc.md](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/content/community/irc.md)
* [https://jb-live.jupiterbroadcasting.net/](https://jb-live.jupiterbroadcasting.net/) - [./config.jblive.toml](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/config.jblive.toml)


While the above PR is totally helpful as-is and should be merged, I wonder if we could also consider the Hugo principle of "everything only exists once" ?

If you think it's worthwhile, I can create a new Issue related to #233 to suggest the idea instead of here..

reclaimingmytime:
@gerbrent Good point! I agree a general goal is to avoid code duplication, in this case, using Hugo variables for the IRC server, channel and default username and extracting the "prefer your own IRC client?" box to a partial. If the server address changes, it only has to be modified once.

Maybe this can be part of JB 2.0.

gerbrent:
Great idea. Created #255 w the new future task..

gerbrent:
Why not use the IRC partial found at https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/themes/jb/layouts/partials/live/irc.html instead of recreating it in https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/content/community/irc.md ?

Issue #265, opened 2022-08-12
host/guest profiles - remove Google Plus links


reclaimingmytime:
The hosts and guests page still links to Google Plus for some hosts and guests, like Joe:
https://jupiterbroadcasting.net/hosts/joe/

The Google Plus link just redirects to the shutdown announcement.

Maybe it's time to remove the link from the template?

gerbrent:
agree!

elreydetoda:
might be a good first issue for someone who's newer and wants to take a little dive into the code :grin: 

Here is where the code is that they'd need to remove:

https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/b716aa253cb6d82990f9f1ee515bf299e8008e16/themes/jb/layouts/partials/people/horizontal.html#L40-L44

It'll remove it for both guests & hosts (which I don't think we have any host that has a google plus link :sweat_smile: )

harriseldon:
I can take this. A good grep on the files and I have a good list. 5 people have a google plus link. Do you want the gplus parameter removed from the json and the follow files changed?
/themes/jb/assets/css/libs/videojs.css
/themes/jb/layouts/partials/hosts/horizontal.html

elreydetoda:
Hey @harriseldon  👋

> good grep on the files and I have a good list

Nice, I didn't even think about doing that 😁 (I'd just thought about the minimum amount of work to "remove it" 😅)

> Do you want the gplus parameter removed from the json

So, I'll default to @gerbrent on this one, but it probably doesn't hurt to just remove all the references to it for all the people.

I will say, we actually just removed all the JSON files from the `data/` folder,  so if you've cloned the repo in the past make sure to do a `git pull` to ensure you've got the most recent changes. (Because you mentioned JSON files)

> the following files changed?

So, typically I don't touch anything in the `themes/jb/assets/css/libs/` folder, in case someone else is using that library for something. The other file should be fine to modify though 😁

Thanks for picking this one up 🥳🚀

(Also, I'd assign you but just found out I can't based on my permission level 😅, which isn't a problem @gerbrent or someone else will be able to add you soon 😁)

harriseldon:
Thanks. I updated my local repository.
This is the command that I am using to remove the gplus lines from all of the md files.
``` shell
find ./ -type f -name "*.md" -exec fgrep -il "gplus" {} \; | xargs sed -i '/gplus:\s\S*/d'
```
[...]

PR #273, opened 2022-08-13
#265 Remove gplus from rendering

harriseldon:
Remove gplus from all of the markdown files for people and removed it from the layout so it will not render


iarwain-forn (approved):
Rest In Peace, Google Plus.

gerbrent (approved):
looks good!

PR #274, opened 2022-08-13
Infra/automated deployment

ironicbadger:
DO NOT MERGE YET

Laying the groundwork for automated production deployments. Comments on methodology welcomed.

The Hugo project does not recommend using the `hugo server` command in production stating that nginx or caddy is their preferred option for these types of deployments. Therefore the Dockerfile was updated to a multi-stage build dropping the usage of the less secure `hugo server -D` command. 

The website, once built in `/site/public` in the build stage, is copied in the nginx container and then docker-compose is used on the production VPS instance to auto detect changes and restart the container only if necessary.

Multi-stage builds mean that the build happens in a separate container (more secure) and the site remains live for as long as possible. Down the road this will also make automatic PR smoke testing more modular and easier to perform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM registry.gitlab.com/pages/hugo/hugo_extended:0.101.0 as builder
-RUN mkdir -p /app
-WORKDIR /app
-COPY . /app
+RUN mkdir -p /site
+WORKDIR /site
+COPY . /site
 RUN hugo
 
 FROM nginx
 RUN rm -rf /usr/share/nginx/html/*
-COPY --from=builder /app/public /usr/share/nginx/html
+COPY --from=builder /site/public /usr/share/nginx/html
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,6 @@ WORKDIR /site
 COPY . /site
 RUN hugo
 
-FROM nginx
+FROM nginx:alpine
 RUN rm -rf /usr/share/nginx/html/*
 COPY --from=builder /site/public /usr/share/nginx/html
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,9 @@
-FROM registry.gitlab.com/pages/hugo/hugo_extended:0.101.0
-CMD [ "serve", "-D", "--bind", "0.0.0.0" ]
-ENTRYPOINT [ "hugo" ]
+FROM registry.gitlab.com/pages/hugo/hugo_extended:0.101.0 as builder
+RUN mkdir -p /app
+WORKDIR /app
+COPY . /app
+RUN hugo
+
+FROM nginx
+RUN rm -rf /usr/share/nginx/html/*
+COPY --from=builder /app/public /usr/share/nginx/html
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,3 +1,9 @@
-FROM registry.gitlab.com/pages/hugo/hugo_extended:0.101.0
-CMD [ "serve", "-D", "--bind", "0.0.0.0" ]
-ENTRYPOINT [ "hugo" ]
+FROM registry.gitlab.com/pages/hugo/hugo_extended:0.101.0 as builder
+RUN mkdir -p /site
+WORKDIR /site
```
elreydetoda, line 3: These can actually be collapsed down to just the `WORKDIR` directive according to the [docs](https://docs.docker.com/engine/reference/builder/#workdir) (it'll auto create the folder if it doesn't exist).

```suggestion
WORKDIR /site
```

Dockerfile
```diff
@@ -1,3 +1,9 @@
...
+COPY . /site
+RUN hugo
+
+FROM nginx
```
elreydetoda, line 7: IIRC [you'd mentioned](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/249#discussion_r940733041) you like the alpine images, is there a reason you're not using that here?

```suggestion
FROM nginx:alpine
```

elreydetoda (commented):
I know you're probably still working on this, but I saw it in my notifications that you'd created this and was excited to take a peek :grin:

Just a small suggestion, question, and wanted to mention:
- we probably should include a .Dockerignore file at minimum for the `.git/` directory

Thank you for doing this :partying_face: :rocket:

ironicbadger:
We already have a `.dockerignore` file [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/.dockerignore) which I thought was adequate.

Thanks for your comments I committed those changes.

ironicbadger:
Need to merge in order to test action (you can't test these in isolation on a branch sadly).

elreydetoda:
> We already have a .dockerignore file here which I thought was adequate.

Lol, my bad it was later at night than I thought 😅

> The Hugo project does not recommend using the hugo server command in production stating that nginx or caddy is their preferred option for these types of deployments.

Ya, @ironicbadger that's absolutely true (for production), but the `hugo serve` command is vital for development workloads. It does hot-reloading (meaning you don't have to restart the container for every change and it'll automatically build only the modified CSS/HTML/pages) and that is an extremely important step for quickly iterating on a problem. Currently the solution you merged *will* require the container to be rebuilt on **every** change.

Also, since there are no more options for the makefile, because we're forcing everyone to use docker for development (from #249) which is fine but means we need to account for development as well, then we either need another Dockerfile (i.e. Dockerfile.dev) which the makefile builds for the dev step or another solution which allows them to use the hugo serve command.

Besides the aforementioned comment I love this PR 😁🥳🚀

elreydetoda:
I made [a PR](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/276) to what I'd mentioned above. Something I did on that Dockerfile.dev that we might want to consider doing on the regular Dockerfile, is I declared an ARG variable which can be passed read during build time.

So, we could keep the hugo version defined in the Makefile and it could be passed to both the production container and dev container to keep things in sync and to ensure we're on the same version for both.

PR #254, opened 2022-08-09
Retrieve Live or Latest PeerTube Episode

iarwain-forn:
### Todo
- [x] Query for the live embed link
- [x] Remove Twitch references
- [x] Call JS from the Live page
  - [x] Automatically update the embed url 🤖 
- [ ] Exclude scheduled episodes from results, or is this a feature 🤔 ?
- [x] Make iframe beauteous
  - [x] Center
  - [x] Embiggen
- [x] Docs
  - [x] jsdocs
  - [x] How to add page specific js

### Testing
- [x] macOS
  - [x] Safari
  - [x] Firefox
- [ ] *nix
  - [ ] NixOS
    - [ ] Gnome Web
    - [ ] Firefox

#122

PR #279, opened 2022-08-14
moving jblive script to head section

elreydetoda:
Meant to catch this during #254, but didn't get a chance to. I remember seeing in the [baseof.html template](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/8d62dd30189cb6e1c311dc23b0b261bb665d78f6/themes/jb/layouts/_default/baseof.html#L4) (like what's done [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/8d62dd30189cb6e1c311dc23b0b261bb665d78f6/themes/jb/layouts/episode/single.html#L1-L3)) you could define something in the head with what I'm doing below. IMO all scripts should stay out of the main body of content (if it can be helped).

So, doing PR to move the JB live script to the head & updating docs on how to do that in the future.


elreydetoda:
tagging @iarwain-forn so they know I'm modifying how their script/docs worked.

iarwain-forn (approved):
That makes a ton of sense! I thought it was a bit odd needing to add scripts mid content if you wanted Hugo to only include it on pages with the partial defined 😅

docs/page-specific-js.md
```diff
@@ -1,26 +1,42 @@
...
+    {{ if hugo.IsProduction }}
+      {{ $fancy = $fancy | minify | fingerprint }}
+    {{ end }}
+    ```
+
+5. Add a `script` tag for the processed JS
+
     ``` html
-    <script src="{{ $fancy.PermaLink }}"></script>
+    <script type="text/javascript" src="{{ $fancy.PermaLink }}" integrity="{{ $fancy.Data.Integrity }}"></script>
```
reclaimingmytime, line 34: `type="text/javascript"` is optional and not necessary when you are using the correct mime type to serve JavaScript.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
```suggestion
    <script src="{{ $fancy.PermaLink }}" integrity="{{ $fancy.Data.Integrity }}"></script>
```

reclaimingmytime:
@elreydetoda 
> IMO all scripts should stay out of the main body of content (if it can be helped).

Can you explain your opinion? ;)

I don't really see a benefit at the moment. When you put JavaScript in the head, it needs to be executed before the page loads. This slows down loading time. The newer approach I've seen is putting JavaScript before the closing body tag so text elements on the page can be shown before the dynamic parts.

docs/page-specific-js.md
elreydetoda: I know it's optional, but is there a reason why it's bad to include it?

I've heard of certain attacks (I don't think it applies to this particular circumstance) that if the server is not specifying the right Content-Type and a user can interact with the server (somehow \*\*waving magical hands\*\*) then they can possibly [get xss](https://portswigger.net/kb/issues/00800400_content-type-incorrectly-stated). Again it's probably not super applicable to this situation, definitely not a web expert at all, but does it hurt anything to have it there?

elreydetoda:
Ya, of course @reclaimingmytime .

So, whenever I go to look at a website and I'm trying to understand how it works. I tend to think that all CSS/JS is defined either in the head or before the closing body tag (like you mentioned).  That way all the actual rendered content is typically in the center of the body tags, and all the dynamic content (JS stuff) is defined outside of the rendered content.

So, while I could've moved it to right before the footer, since there wasn't a convention to do the later in the baseof.html template and it is already being done in the [episode/single.html template](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/bcf87f98521dfc74c1a0c5897943e24b30af01c3/themes/jb/layouts/episode/single.html) I was just following precedence.

From the slowdown perspective, since it's an async function doesn't it just asynchronously execute in the background after the window.onload event? Or would it be better to throw a `defer` attribute into the script declaration?

docs/page-specific-js.md
reclaimingmytime: I guess it doesn't hurt, besides adding a few more bytes to the page download. 😉

reclaimingmytime:
> From the slowdown perspective, since it's an async function doesn't it just asynchronously execute in the background after the window.onload event? 

I think that's regarding when the script gets executed. Part of the script still gets downloaded and evaluated when it's in the head.

> Or would it be better to throw a defer attribute into the script declaration?

I haven't used `async` or `defer` as attributes for the `script` tag before, so I can't say if that's a usable alternative.

It shouldn't make too much of a difference in practice though, as there's still the latency of fetching the PeerTube API, and the JavaScript only gets fetched once and is then cached.

elreydetoda:
> It shouldn't make too much of a difference in practice though, as there's still the latency of fetching the PeerTube API, and the JavaScript only gets fetched once and is then cached.

So, @reclaimingmytime do you think it's ok as is, should we add a construct to add it to the footer, or use async/defer?

Like you said, I don't think it really matters that much overall for this situation (since it's so small) but just wanted to get an explicit response to the situation. :slightly_smiling_face:

reclaimingmytime:
@elreydetoda Yes, all things considered, I think this pull request is fine as it is 😄

elreydetoda:
> @elreydetoda Yes, all things considered, I think this pull request is fine as it is smile

It was a fun theoretical convo though :grin:

ironicbadger:
So this wants a merge?

elreydetoda:
> So this wants a merge?

Yes, please merge.

Issue #281, opened 2022-08-14
Jupiter.tube embed size - larger please!


gerbrent:
It would be great if the Peertube embed could have a larger size. Currently feels quite small on a desktop monitor. Something akin to the default Youtube player might be a good starting point to decide on the default desktop size..

As example, here's a size-by-size with Youtube.

![image](https://user-images.githubusercontent.com/7018928/184557527-32066d1b-ec73-474c-9580-1b4e87790139.png)

noblepayne:
@ChrisLAS suggested 960x540.

noblepayne:
We may want a responsive or mobile-friendly version in the future, but created #282 to get things a little bigger in the meantime.

elreydetoda:
Related #278

Maybe that can be the responsive issue

PR #283, opened 2022-08-15
making jb-live more responsive

### LINKED ISSUE #278: Adjust PeerTube iframe width on mobile devices
The width of the dynamic PeerTube iframe is wider than the device width on mobile devices, so the page scrolls horizontally.

Screenshot_20220814-154919

Screenshot_20220814-154930

elreydetoda:
close #278 
related #281 

I'd love to have maybe @gerbrent, @noblepayne (since he did #282), @reesericci (since opinion was asked in #278), or @reclaimingmytime (created #278)  and check to see if this size works for you all on both desktop & mobile.

You can test this locally by using [dev tools](https://developer.chrome.com/docs/devtools/device-mode/) (it's pretty much the same in firefox), or binding to all interfaces (`--bind 0.0.0.0` - use caution and make sure you're on a trusted network, because this will expose a port up to the local network) and navigating to your laptop's IP address.

### DIFF
--- a/themes/jb/layouts/partials/live/jb-tube.html
+++ b/themes/jb/layouts/partials/live/jb-tube.html
@@ -1,8 +1,8 @@
 <div class="jb-tube-wrapper">
   <div class="jb-tube" style="text-align: center;">
-    <div class="jb-tube-video">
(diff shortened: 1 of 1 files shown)


elreydetoda:
Closed via #317

Issue #46, opened 2022-07-07
hugo WARN: no layout file for "webplayer" for kind "page"


gerbrent:
solve hugo runtime WARN:
> found no layout file for "webplayer" for kind "page": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.

gerbrent:
![image](https://user-images.githubusercontent.com/7018928/177740312-c370d376-4a45-482f-833f-788e027292fc.png)

StefanS-O:
I think those should only be generated for kind episode

kbondarev:
I searched around the web for those warning messages and that lead me to the config.toml file and [this hugo doc page about output formats](https://gohugo.io/templates/output-formats/). 

Changing these lines at the bottom of config.toml by deleting bunch of stuff stops the warnings:

> Original:
> ```toml
> [outputFormats]
>   [outputFormats.webplayer]
>   mediatype ="application/json"
>   baseName = "webplayer"
> 
> [outputs]
>   page = ["json","html","webplayer"]
>   section = ["html","rss"]
> ```
> 
> Leave only this:
> ```toml
> [outputs]
>   page = ["html"]
>   section = ["html","rss"]
> ```
> 


@StefanS-O, can help confirm my guessings here:
- The "json" output for `page` would make basically a JSON API for each episode? But it spits the warning since there's no template?
- The "[outputFormats.webplayer]" is a special definiton of an output format for the Podlove player? To consume things like chapters, related episodes, etc? (I think you mentioned something about it before)

elreydetoda:
no more :slightly_smiling_face: 
![image](https://user-images.githubusercontent.com/10230166/185450765-e52b1714-4919-4373-a6a1-ec34c12ed44f.png)

PR #290, opened 2022-08-16
fixing the warnings that show during deployment

elreydetoda:
apparently these page types were causing all the warnings to show up during development and deployment...I don't see why they're necessary?

image

@StefanS-O , were those there for some reason?

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -220,6 +220,7 @@ paginate = 12
   baseName = "webplayer"
 
 [outputs]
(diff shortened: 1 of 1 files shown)


gerbrent:
fixes #46

StefanS-O:
it was stuff we needed for the web player we had before. it needed the JSON output for config per episode. I also generated JSON files for each episode, but that is also not needed anymore

StefanS-O (approved):

PR #298, opened 2022-08-16
[people] update danjohansen.md

Strit:
Added my LinkedIn link.


gerbrent:
great!

need to make sure changes such as these dont get overwritten, and i believe @kbondarev has a method there.

Issue #294, opened 2022-08-16
Shorthand redirects off main website


elreydetoda:
Mentioned shorthand redirectors off of main website during Office Hours 10:
- /live to https://jb-live.jupiterbroadcasting.net/
- /mumble to https://new.jupiterbroadcasting.com/community/mumble/

@ChrisLAS had mentioned he'd possibly have more as well.

ironicbadger:
- [x] matrix
- [ ] ...?

kd2flz:
yeah, just noticed this. The "live" link still redirect to https://jb-live.jupiterbroadcasting.net/

gerbrent:
see #161 for a slightly more complete list...

gerbrent:
> yeah, just noticed this. The "live" link still redirect to https://jb-live.jupiterbroadcasting.net/

I really wish someone could do #233 that would likely solve some of these.... (maybe I should learn how ; )

kd2flz:
I'll look at #233 this evening. Not sure if it's within my skillset, but worth a shot for sure.

elreydetoda:
> > yeah, just noticed this. The "live" link still redirect to https://jb-live.jupiterbroadcasting.net/
> 
> I really wish someone could do #233 that would likely solve some of these.... (maybe I should learn how ; ) Will make a new ticket for the above, and I'll likely get started on solving it too (trivial...)

@gerbrent I had a question I'd posted on that issue (#233), to clarify what you were wanting from it, and I was going to do it so I could also just close #300

gerbrent:
oh sorry! question answered (hopefully satisfactorily) 

@kd2flz: ok to let @elreydetoda take #233 since he's hugo-skilled, which will likely be essential on this one?

PR #300, opened 2022-08-16
redirector for /live to jb-live domain

elreydetoda:
unfortunately there doesn't appear to be a native way to redirect to a url that's not hugo based for the current site (i.e. jb-live.jb.net is technically a completely different deployment)

Normally if we'd redirect to another internal page to the repo it'd be easy (see #297 & #295), but since it's referencing an external site it not as easy (look at #294 for more info). 

So, the only way I can think to handle this is by defining a static file (which doesn't get access to hugo variables (I tried to reference it), pic below), and having the redirect there manually defined. (based on hugo's default page when calling an `alias`)

(what happened when trying to reference a hugo variable)
image

### DIFF
--- a/static/live/index.html
+++ b/static/live/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en-us">
+
+<head>
+    <title>https://jb-live.jupiterbroadcasting.net</title>
+    <link rel="canonical" href="https://jb-live.jupiterbroadcasting.net">
+    <meta name="robots" content="noindex">
+    <meta charset="utf-8">
+    <meta http-equiv="refresh" content="0; url=https://jb-live.jupiterbroadcasting.net">
+</head>
+
+</html>


gerbrent:
sure would be nice to solve #233 if it simplifies this PR/functionality for you...

RealOrangeOne:
This redirect is probably better done with nginx than like this. It would require vendoring our own nginx config, but that's probably a good idea anyway for extra customization.

elreydetoda:
> This redirect is probably better done with nginx than like this. It would require vendoring our own nginx config, but that's probably a good idea anyway for extra customization.

So, I actually just copy/pasted the exact redirect page that hugo generates with aliases: https://gohugo.io/content-management/urls#aliases and swapped out the values we wanted.

Also, IIRC @ironicbadger mentioned we're using traefik so it'd be a config change (probably just a label) for that. I was going to post an issue on the infra repo, but figured I'd draft it here and get opinions.

To @gerbrent's point though, if #233 was taken care of I'd just be able to add an alias to it just like the other ones.

RealOrangeOne:
Nginx is running in the container, so it's still possible to do it there, regardless of what sits in front of it.

elreydetoda:
Closed via #317

PR #303, opened 2022-08-17
adding back thing removed from merge conflict

elreydetoda:
sorry, when trying to remove the merge conflict I was looking at the wrong place when I was validating functionality...we need this :sweat_smile:


ironicbadger:
what is this?

elreydetoda:
During a merge conflict I removed the show listings by accident.

image


I'd removed some duplicate code from another file before and thought it was that file during the merge conflict resolution.

image

PR #304, opened 2022-08-17
forgot to remove draft...

elreydetoda:


ironicbadger:
what is this?

elreydetoda:
I'd run hugo in development mode (as I normally do) when creating the subscribe.md files, but this time I used the hugo new command (instead of creating them from scratch) to create the files. By default that command uses the default archtype, which sets the draft status of the markdown file to true which when we deploy to prod ignores those files (since they're in draft mode). So, that's why the coder /subscribe file isn't there in production. And this PR is changing all those draft status to false, so it can be deployed to prod properly...sorry :disappointed:

ironicbadger:
Please a create a new issue to track this as a test we need to write for our CI.

PR #335, opened 2022-08-25
Bug/boost - typos, spelling, grammar, improvements, additions

gerbrent:


elreydetoda:
We should probably mention the 2000 sats thing here, right? 
That also let's you give a place you can explain why and then a consistent location you can tell people to go to when they ask why. (Just thought about it since you were editing the boosts page)

gerbrent:
Good thought! But... that's likely changing somewhat frequently as we find a nice balance... and not necessary to define here, I don't feel.

PR #325, opened 2022-08-20
Setup End to End Tests

kylepotts:
Should help start the ball following on solving https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/286#issuecomment-1215477732

PR #317, opened 2022-08-19
Migrating JB-live to main site /live

elreydetoda:
closes #309
closes #233 
closes #278
closes PRs #300 & #283
related #294, #163 

This does the following:
- migrates the jb-live.jb.net separate config/site :arrow_right: new.jb.com/live
- removes config file for the extra site being built
- changes the menu entry of "Live" to point to the new page
- removes the params which referenced the live URL
- adds more responsiveness
- deletes the GH action build step for jb-live deployments

PR #353, opened 2022-08-29
E2E GH Action - patch/modifying

elreydetoda:
So, I think there was an issue with this action: https://github.com/jakejarvis/hugo-build-action

Because the prod deployment works and after I swapped over to use the [same docker image](https://github.com/elreydetoda/jupiterbroadcasting.com/commit/867840d1c3fcf479caa613eff09d909f555dbc37#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR13) as the [prod deployment](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/2fe5a3969c1a36cb4fc145a6aa080daeb2b04837/Dockerfile#L1) everything worked fine. (from the hugo building perspective)

After that I adapted the action to properly use the pipenv (Pipfile + .lock) instead of the requirements.txt.

Next the tests were actually working and failing as expected! :joy: It was failing because the JB live link wasn't as [expected from the tests](https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8080911236?check_suite_focus=true#step:9:48), and because #317 was merged before #325 was merged (and #325 wasn't rebase'd on main) the tests were expecting the old URL. So I updated the reference to point to `/live` instead, and everything works now :grin: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8080987260?check_suite_focus=true


gerbrent:
great 🔍 investigative work @elreydetoda ! I could tell it was working as expected - i.e. the E2E tests were finding a problem as configured, but didn't know how to inspect.... hope someone teaches me that at some point ; )

elreydetoda:
> great mag investigative work @elreydetoda ! I could tell it was working as expected - i.e. the E2E tests were finding a problem as configured, but didn't know how to inspect.... hope someone teaches me that at some point ; )

From a high level, this was pretty much my process:

1. I identified an error because of the error starting here: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8080911236?check_suite_focus=true#step:9:48
2. Then the line below it here says the actual value: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8080911236?check_suite_focus=true#step:9:49
3. So, I knew that the comparison of expected vs actual was not right.
4. You can see the expected dropdowns passed in here: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8080911236?check_suite_focus=true#step:9:29, and can see where it's got the jb-live.jb.net domain. 
5. I went into conftest.py file, which is used to be a central location for passing in commonly shared value (i.e. all the dropdowns we're expecting, because that might be used by multiple tests), and changed it from the jb-live.jb.net to `/live`: https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2R59

gerbrent:
I don't understand the implications of `pipenv` enough (though aware we've implemented it elsewhere recently) so asking for reviewers:

elreydetoda:
> I don't understand the implications of `pipenv` enough (though aware we've implemented it elsewhere recently) so asking for reviewers:

Definitely feel free to still wait for the reviews, but just for some extra info this is what I mentioned to @kbondarev for why we should use pipenv: https://github.com/JupiterBroadcasting/show-scraper/issues/11

elreydetoda:
Converted to draft, because of weird bug that came up with the E2E tests: https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/383#issuecomment-1235832886

Going to redo how the E2E GH action works to hopefully make more consistent and have less quirks...

gerbrent:
....did we find you in a rabbit hole?? ; )

.github/workflows/build_container-prod.yml
```diff
@@ -0,0 +1,64 @@
+name: Building Prod Container
```
gerbrent, line 1: #386

elreydetoda:
> ....did we find you in a rabbit hole?? ; )

That's so funny you said those word! Look at what the title of my blog is: https://elrey.casa/blog (it's "The Rabbit Hole") 🤣

elreydetoda:
alright... I think I've finally gotten it :upside_down_face: 

Going to make a longer comment to explain things, but this will be what the final report will look like after merging: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8199597176

elreydetoda:
putting back to draft till I have time to finish documenting things and make my long comment (almost ready :sweat_smile: )

elreydetoda:
I re-worked some more things, and with my [latest commit](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/commit/4be84faa387e79f9a76641e0edbe6f9eca1fd208) we shouldn't get failed tests from people submitting PRs, because it'll skip the upload reports thing. You can still go into the action and see the test results (plus it'll fail if the tests fail), but I think it keeps telling us that it's failing because they don't have permissions to write to the repo or something like that.

elreydetoda:
alright...so this PR started because I was simply going to refactor the E2E (end to end) tests a bit (talked about that with my initial commit [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353#issue-1354940804))...but it evolved into much more than that :sweat_smile: 

---

### GH actions

I'm still accomplishing everything from the original comment, but because of #383 not working in the GH action CI (even though it worked locally) I tried to make things more reproducible with local testing & the GH action CI testing. This change was what caused me to create the 2 GH actions to build containers via the CI (with some [accompanying documentation](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-946785ae098a43c945b0386c91a1553c6ac8c857d521658e157af47888e3ef2d) to setup credentials for pushing to Quay): 

- [production container](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-c3d53273f274f7cd99d5ef9a6f7bdcb27fb62a67f6d62700bef49b0a9dde420a)
  - I'm actually building 2 container in parallel for this action, and mainly pushing them with the labels of (read the GH action notes to see how I'm double tagging them):
    - tests_web-latest: this is the latest version of production which'll be exactly what gets used when the tests run against their docker container
      - this isn't the actual container the tests use, because of the following reasons:
        - it would take way longer for a contributor to get feedback if that's what we based it off of
          - the action would have to complete for building
          - then it'd kickoff the action for the tests which would pull down the image it just created and run the tests against that
        - it'd be overwritten if we had multiple PRs open at the same time
        - it'd allow people contributing to be feeding code directly into a JB container, which could be a security risk
    - prod-latest
      - this isn't actually getting used yet, we'd have to chat with @ironicbadger to see what he feels about us implementing this
        - before @ironicbadger [mentioned](https://matrix.to/#/!rDlEtonuSythLzmxQL:jupiterbroadcasting.com/$BkD3XySMzqQWgy8gM0FIdYgWACM3JyPXXj9K7HawGt4?via=matrix.org&via=jupiterbroadcasting.com&via=mooreltd.net) he wasn't averse to the idea, but just would have to needed more info
      - I've also included a commented out trigger we could use for the [deploy to prod](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-4ce342fb97e7d7feb8a578c91ba89fe4c27ddf16b4306909fe6d90220141913dR16)
        - I'd say lets not implement that with this PR though, so we can get tests up to par first
      - this is also not being used because of all the reasons listed above for the other container
- [tests container](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-d2953aa3ef4a677ead43036f8b9e23cfc796c70349831ddae1181c5b7175b363)
  - this container will only rebuild when there are dependency updates or it's dockerfile gets modified.

[Here](https://quay.io/repository/elrey741/jb_web_container) is a link to my quay container images, so you can see what the tags are and how it'll look in the JB quay.

---

### Tests CI run environment

The actual tests (to prevent #383 in the future) are [run](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR32) with a [docker-compose file](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-3bea5d63cb5af722ee3674771d619ad25fbd70d6ebd67ad537156f1d4bf86ccb), and this compose file uses the Dockerfiles that are in the repo to build the containers for testing. This gets us a few things:

- allows us to pull in all local changes for PRs without introducing security risks into our GH action environment
- allows the e2e GH action to be self-containing (not have dependencies on other actions or secrets for the repo)
- lets us have a repeatable environment for testing locally vs testing in the GH action pipline (since they're both using the same dockerfile)
- lets us run the docker-compose stack locally to debug if there's an issue with the CI

I tried using [GH service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) but it wouldn't let me re-use the local dockerfiles (which wouldn't allow us to leverage everything I listed above)

---

### E2E test modifications

So, the only thing I could think playwright might've been failing was that when the JB well known file is access it doesn't load up a DOM or any HTML elements at all. So, I'm thinking that the playwright auto-wait feature was waiting for the [load event](https://playwright.dev/python/docs/api/class-page#page-wait-for-load-state) never triggered or something. So, I knew that Playwright also supported [APi testing](https://playwright.dev/python/docs/api-testing), and the response we're getting from the well_known file was similar to what we'd get from an API response. So, I created an [API context](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2R79-R96) and used that to then handle the [retrieval of the well known file](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-d38428a14b3e2789cec95f3e6f87541a154ccb76bf7b93a48203fbebae6f2571R8)...and everything worked...both locally and in the CI.

#### future E2E tests

Since I was making that modification and cleaning things up a bit I also modified the [test_home.py file](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-cc8322b4267814b800cef19b4cd432af2a990aad2f5f63dbc281e40962f6bedd) as well:
- I removed the `.pause()` call (mentioned it [wasn't necessary](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/325#discussion_r953135829))
- I also removed all the `page.goto("/")`, and made a fixture to auto be used when all tests in that file is run (making the code more DRY, and not requiring them to each have a `goto` line to navigate to the home page
- Lastly, the `test_dropdown` function was using different nested elements as targets for checking the dropdowns (other functions are still using this as well, but I'm choosing this one as an example on how other would be refactored as well)
  - so instead I'm not [architecturing the conftest](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2R28-R52) to more resemble the structure of the dropdowns
  - that structure change lets us tell playwright to act more like a user
    - hover over the menu option
    - then check each link based on the text the user would see and make sure it's visible to the user

:arrow_up:  is something that we'll need to keep in mind and document on how to create test in the future, but I just wanted to modify this one so it could be an example for others.

gerbrent:
whoa.

WHOA! @elreydetoda ! This is impressive...

Going to bring in a few folks on reviews and testing.

gerbrent:
@kylepotts - care to have a look at the Playwright aspect to help review this?

test/conftest.py
```diff
@@ -69,7 +69,9 @@ def expect_nav_items() -> List[Dict[str,str]]:
         {'title': 'Boost!', 'href': '/boost'},
         {'title': 'Garage Sale', 'href': 'https://www.jupitergarage.com/'},
         {'title': 'Membership', 'href': '/membership'},
-        {'title': 'Archive', 'href': '/archive'},
```
gerbrent, line 64: nice catch. We still need to put the E2E in our mental image of all PRs and features and changes.....

Reference: addresses failed E2E based on change in #399

elreydetoda:
BTW, I merged instead of rebasing so that all my links to stuff from my [comment above](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/#issuecomment-1240682632) didn't break things.

gerbrent:
Hey @elreydetoda - any thoughts on the best way to test this for us mere mortals?

I wonder if you'll say "a dev branch!" to which I would say LETS DO IT!! 💪🏻

elreydetoda:
Lol, ya that'd probably be the easiest.

If that's the case though I might need to make a modification or two to the PR (so it properly triggers on the develop branch as well as main). I'll make that modification tonight, and I'll let you know when it's good to go.

I'll also need to change my target branch for the PR (since it's currently targeting main).

elreydetoda:
Alright, @gerbrent I've created the develop branch, and now I'm pointing my PR to that branch. You'll need to still follow this [accompanying documentation](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353/files#diff-946785ae098a43c945b0386c91a1553c6ac8c857d521658e157af47888e3ef2d), so the containers build properly (or you could disable those actions from running if you want for now).

Here's what the workflow will look like:

- PR into develop - https://github.com/elreydetoda/jupiterbroadcasting.com/pull/13
  - Merge into develop:
    - develop containers being built: https://github.com/elreydetoda/jupiterbroadcasting.com/actions/runs/3042253566
    - E2E tests: https://github.com/elreydetoda/jupiterbroadcasting.com/actions/runs/3042253568
      - report: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8320186125
- PR into prod from develop: https://github.com/elreydetoda/jupiterbroadcasting.com/pull/14
  - Merge into develop:
    - prod containers being built: https://github.com/elreydetoda/jupiterbroadcasting.com/actions/runs/3042276766
    - E2E tests: https://github.com/elreydetoda/jupiterbroadcasting.com/actions/runs/3042276791
      - report: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8320252541

elreydetoda:
BTW @gerbrent , if you want we can hop on a call and walk through the setup of this environment as well 😁 (whatever you're most comfortable with 🙂)

gerbrent:
Merging this as Part A, with Part B involving some JB/ @ironicbadger involvement to handle deployment functionality. We will also need to sign up to quay.io in that process, so definitely more JB Action Required.

Onward!

gerbrent:
@ironicbadger - care to link up w @elreydetoda to complete the final stages of this PR?

PR #369, opened 2022-09-01
GH workflow - remove redundant `git fetch`

gerbrent:
this file features a `git fetch && git pull` - this is rather redundant, considering the [Git Documentation](https://www.git-scm.com/docs/git-pull):

> git pull runs `git fetch` with the given parameters and then depending on configuration options or command line flags, will call either `git rebase` or `git merge` to reconcile diverging branches.

so, while very minor, we can I believe safely drop the redundant `git fetch`

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -21,7 +21,7 @@ jobs:
             USERNAME: ${{ secrets.VPS_DEPLOY_USER }}
             PORT: ${{ secrets.VPS_SSH_PORT }}
             KEY: ${{ secrets.VPS_SSHKEY }}
-            script: cd ${{ secrets.VPS_PROJECT_PATH }} && git fetch && git pull
+            script: cd ${{ secrets.VPS_PROJECT_PATH }} && git pull
             
       - name: Deploy
         uses: appleboy/ssh-action@master


elreydetoda (approved):
Seems reasonable to me 😁

I didn't even notice that 😅, nice catch @gerbrent 🥳

ironicbadger:
There was a reason I did both. I don't recall why now though.

gerbrent:
@elreydetoda - would you be willing to test this one? That way either we can deny or confirm @ironicbadger 's memory ; )
Since you're a temporary infrastructure 🥷 I figured this would be relatively easy for you to investigate.

low priority, of course.

elreydetoda:
Sure, I can check it out this afternoon 😁

elreydetoda:
So...I honestly have no idea why fetch would be needed.

From my understanding of it (based on [this](https://www.git-tower.com/learn/git/faq/difference-between-git-fetch-git-pull/)), it will only get new information. From an automation perspective, since there's not going to be any human interpretation of this new information (i.e. don't do a git pull if there's going to be a merge conflict), then I see no reason why it'd need to be in there.

the fetch doesn't modify your local code at all, and I remember vscode offers to do a [periodic fetch for you](https://code.visualstudio.com/Docs/editor/versioncontrol#_remotes), because it won't harm your local repo data, to show you possible syncing you'd need to do:
> VS Code is able to periodically fetch changes from your remotes. This enables VS Code to show how many changes your local repository is ahead or behind the remote.

So, I don't see a reason why it'd be needed at all (in this circumstance). I think I might have heard before that it's used to pull updated tags (and git pull doesn't by default or something), but since we're not tagging anything I don't think that matters to us.

gerbrent:
thanks for also working your way through the logic @elreydetoda  - I'll merge and note to us that we should look for related issues..

gerbrent:
seems to be working just fine: https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/runs/8232673539?check_suite_focus=true#step:4:20

PR #370, opened 2022-09-01
E2E - update Live menu item check address

gerbrent:
update Live menu item check address from `https://jb-live.jupiterbroadcasting.net/` to `/live`

PR #373, opened 2022-09-01
Feature/update host alex

ironicbadger:
updates my host image and brents website url.

built and tested locally.


elreydetoda:
Just wanted to note:
* will probably be a good idea to add Alex's name to the dontoverride variable here: https://github.com/JupiterBroadcasting/show-scraper/blob/main/config.yml#L61
* failing test isn't from this PR see #370 & #353

gerbrent (approved):

PR #374, opened 2022-09-01
add kate project files

gerbrent:


elreydetoda (approved):
I've never had these pop up before, but sure why not :upside_down_face:

gerbrent:
.... probably cause I'm the only one using Kate and exploring it's features ; )

elreydetoda:
I actually use kate, but just not as my code editor. So, yes to you exploring more of it's features :joy:

Issue #381, opened 2022-09-02
🚨 ⚠️  Matrix Federation Broken due to missing .well-known dirctory and contents. JB chat down 🚨 ⚠️ 


ChrisLAS:
Seems we broke Matrix federation when we switched over to the new site. Lurking in the file depths of our old site was a file and folder structure that verified our server to the Matrix federation. 

Some more info here: https://matrix.org/docs/spec/server_server/r0.1.0.html#resolving-server-names

I've grabbed our old .well-known folder with its files. But I have NO idea how something like that gets deployed in this setup. In our previous setup, that lived in the root web directory.

But we need this done ASAP, since it's been since Tuesday evening our Matrix service has been down for the community and is impacting show production directly. 🚨 

My snag of the old files: https://files.horse/s/YPKJfJQ6XiKQMLk

noblepayne:
Attempted to add our previous well-known delegation setup as a static asset.

ChrisLAS:
Quick fix with static assets did NOT work, re-opening this issue.

noblepayne:
Looks like it WAS working for the `www`  domain, but not the root. I've updated the cloudflare page rule that is responsible for that forwarding to pass through `/*`, and at least according the [federation tester](https://federationtester.matrix.org/#jupiterbroadcasting.com) looks like we are back in business.

elreydetoda:
@noblepayne , did you update it with `$1` like how I'd screenshot'ed in [this comment](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/358#issuecomment-1233341553)?

Just want to know so we can update this PR: https://github.com/JupiterBroadcasting/infra/pull/9

It'll make sure terraform doesn't wipe out the manual change in future runs :sweat_smile: 

Also, we should integrate an E2E test (and maybe uptime kuma (but that might be overkill)) to make sure that all future deployments have this file in it with the expected response (the json blob from #382). We **definitely** don't want to break the matrix configuration again... :sweat_smile:

noblepayne:
@elreydetoda 
> Also, we should integrate an E2E test

Saw your work on this, thank you!

> Also might want to consider doing some type of periodic check on the federation tester 

Was thinking the same! :brain: 

> did you update it with $1 like how I'd screenshot'ed

Yup!

> Just want to know so we can update this PR

Updated :+1:

PR #383, opened 2022-09-02
adding test for checking matrix well_known file

elreydetoda:
should prevent #381 from happening again.

### DIFF
--- a/test/e2e/test_infra_configs.py
+++ b/test/e2e/test_infra_configs.py
@@ -1,0 +1,14 @@
+from json import loads as j_loads
+from playwright.sync_api import Page
+
+def test_matrix_well_known(page: Page):
+    contents = page.goto("/.well-known/matrix/server").text()
+    
+    well_known_structure = {
+        "m.server": "colony.jupiterbroadcasting.com:443"
+    }
+
+    if j_loads(contents) == well_known_structure:
+        assert True
+    else:
+        assert False


gerbrent:
nice one!

gerbrent:
helps avoid #381

gerbrent:
looks like some E2E errors. I'll check the assumed path in the test of `/.well-known/matrix/server`

image


https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/runs/8162128923

But there is also another failed test that might relate..

gerbrent:
/.well-known/matrix/server is good:

`curl -I https://www.jupiterbroadcasting.com/.well-known/matrix/server
HTTP/2 200 `

elreydetoda:
Ya...I was just noticing that (I'd expected the E2E tests to run when I submitted my PR, but #385 says why it didn't).

For some reason it looks like it's [not passing the page object right](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/runs/8162095907?check_suite_focus=true#step:9:78): `<Page url='about:blank'>`

This is what it [should be passing](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/runs/8162095907?check_suite_focus=true#step:9:32): `<Page url='http://localhost:1313/'>`

I don't know why this is happening yet, but everything ran perfectly fine in the local testing environment... :thinking: (still figuring out why it's not working)

[asciicast](https://asciinema.org/a/iVKZ3jCxNFp5hl4cqQ1ND7aSp)

elreydetoda:
I'm even more confused now... 🤦

So...I was doing some testing, and I created 2 more branches off of my #353 PR...

1. was no modification besides adding the new check I'd created & adding the new well known from #382 
2. I pulled all the changes from main and then re-applied my commits with #353 (so it'd effectively be what main would look like after #353 is merged)

Number 1 failed with the same error we're seeing with this PR: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8163979470?check_suite_focus=true#step:9:28
Number 2 failed with the same error we're seeing with this PR: https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8163879397?check_suite_focus=true#step:9:28

but then things get interesting...I removed the well known file commit to my Number 1 branch, and while it still failed (because the well known file no longer exists) it actually passed in the right object (with the base_domain)!!! :thinking: 
https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8164035325?check_suite_focus=true#step:9:28

gerbrent:
so confusing!! hmmmmm..............

gerbrent:
@elreydetoda and I just spent much time trying to solve this one together, and..... 🥁 drumroll! 🥁 ........ didn't solve it! BLAST.

elreydetoda:
alright...I think I'm starting down the right path to solve this...I've finally been able to replicate it! (locally & in GH actions)

https://github.com/elreydetoda/jupiterbroadcasting.com/runs/8180359800?check_suite_focus=true

(ignore the other error, that's because of the way I'm currently building the container and I'm about to change it to fix that :upside_down_face: :sweat_smile: )

The only consistent change between these situations is that it's not  the hugo binary serving these files...when we're doing the Dockerfile.dev (make dev) locally we're using the hugo file to load the file (which serves it from memory), but when I'm serving it via the container (through nginx or `python3 -m http.server`) I get the error...

I also found that some other people might be having a similar issue with the pytest-playwright library: https://stackoverflow.com/questions/68266451/navigating-to-url-waiting-until-load-python-playwright-issue

I tried their quick fix solution, but it didn't work so still investigating 🧐

gerbrent:
sounds like a potential bug in pytest's playwright library perhaps??

Issue #371, opened 2022-09-01
deploy after scraping - not deploying despite successful scrape


gerbrent:
Noticed a behaviour:

There has been a successful "scrape & commit" action on Coder 481 [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/commit/1e4fac76e1a10a55edcf1f429f3507de69a5e13f) about "17 hours ago"

Great! (despite other unrelated issues  #367 )

This, however, is not being pushed into production at time of scrape, and thus [Coder 481](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/content/show/coder-radio/0481.md) exists in code, but has not been published to jupiterbroadcasting.com. Should be https://www.jupiterbroadcasting.com/show/coder-radio/481/ - currently a 404.

### Questions
* Do I have this right? 
* Is this by design, then, and we just haven't noticed w all the feature PRs being pushed, publishing newly scraped episodes along w PR merges?
* Should we be pushing scraped data on successful scrape?
   * this can be dangerous.. if the scraper misbehaves, we do not want to wipe production data..
   * so, perhaps our new E2E tests can come into play? If tests pass, then publish. If not.... do not.
      * may need to write new tests to specifically target known-good episodes exist..

gerbrent:
I will do a manual deploy to prod via workflow run to publish Coder 481.

gerbrent:
[Deployed](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/actions/runs/2969089729) and Coder 481 now published 👍🏻

gerbrent:
related: #215 #365 #366 #367 (possible duplicate)

elreydetoda:
> Should we be pushing scraped data on successful scrape?

IMO, yes we should.

> this can be dangerous.. if the scraper misbehaves, we do not want to wipe production data..

in theory there shouldn't be that much to worry about, because it could easily be disabled via the web-ui and someone can revert the commit. Since the bot isn't doing a force push, which could erase history, it should then trigger a re-deploy after the commit being reverted.

If we had a versioned artifact to pull from (summarized [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/307#issuecomment-1234139769)), then this would be an even less concerning situation (I never got feedback though on if we should proceed with that or not though :sweat_smile:).

[...]

PR #385, opened 2022-09-02
allowing PRs to run E2E tests

elreydetoda:
So...I've typically been disabling the different CI/CD test that we created on my personal repo. Most of them were deployment things which weren't going to work on my account and then I disabled E2E test because they were failing.

But after disabling E2E tests, and I'd make [a PR](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/383)...the E2E tests didn't run for my PR. Then I looked at the config again and it's only running on push'es.

https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/d141be3dcfefd101185cefdb814dc567298d91e5/.github/workflows/e2e.yml#L3

Which means when people do PRs it won't require it to run from JB's repo when someone does a pull_request.

This'll allow tests to not run if they disable it (like what I did), and so I think we should force the tests to run when people submit PRs (also keeping the push to main till we figure out #371).

Here is an example of it working after I merged it with my main on my fork (check under the all tests section near the bottom): https://github.com/elreydetoda/jupiterbroadcasting.com/pull/7

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,6 +1,16 @@
 name: E2E Tests
 
-on: [push]
+on:
+  # keeping this till we figure out #371, so we can at least validate the
+  #   scrapper PRs aren't breaking things
+  push:
+    branches: [ main ]
+
+  # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
+  pull_request:
+
+  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


gerbrent (approved):

gerbrent:
#371 is satisfied, so this looks fine to me.

Will have to keep an eye on it to make sure running as intended.

PR #396, opened 2022-09-06
:technologist: makefile - add `dev-build-metrics`: hugo build metrics reporting

gerbrent:
Added the helpful build metrics for performance reporting, if so desired. 
I'm happy to be told this isn't needed here..

PR #406, opened 2022-09-08
making Makefile code a bit more DRY

elreydetoda:
related to #396

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
-dev:
+dev-build:
 	docker build --rm -f Dockerfile.dev -t jb_hugo:latest .
+
+dev: dev-build
 	docker run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest
 
-dev-build-metrics:
-	docker build --rm -f Dockerfile.dev -t jb_hugo:latest .
+dev-build-metrics: dev-build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,9 +1,10 @@
-dev:
+dev-build:
 	docker build --rm -f Dockerfile.dev -t jb_hugo:latest .
```
elreydetoda, line 2: pulling out the build definition from each of the subsequent dev definitions allows you to re-use this step in other places.

Makefile
```diff
@@ -1,9 +1,10 @@
...
+
+dev: dev-build
```
elreydetoda, line 4: this tells make, if you're going to do this step then `dev-build` is a dependency of this step.

Makefile
```diff
@@ -1,9 +1,10 @@
...
 	docker run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest
 
-dev-build-metrics:
-	docker build --rm -f Dockerfile.dev -t jb_hugo:latest .
+dev-build-metrics: dev-build
```
elreydetoda, line 7: (same thing as above) this tells make, if you're going to do this step then `dev-build` is a dependency of this step.

gerbrent: nice!

gerbrent (approved):
tested and confirmed working as intended 👍🏻

PR #411, opened 2022-09-08
Check if exists .Title and add pipe

### LINKED ISSUE #405: Home - `|` separator appears on tabs/title bar despite not being needed/useful
When on the home page, the `|` title separator appears despite not being needed/useful. It is, however, useful in other contexts.

The task: Show the `|` separator in page title only when needed.

image

Other page where it's useful:

image

ChanceM:
Adds conditional to see if a title is present and add the pipe as needed.

Fixes #405

### DIFF
--- a/themes/jb/layouts/partials/meta.html
+++ b/themes/jb/layouts/partials/meta.html
@@ -1,9 +1,9 @@
 <!-- Meta Ttitles -->
-<title itemprop="name">{{ .Title }} | {{ .Site.Title }}</title>
-<meta property="og:title" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta name="twitter:title" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta itemprop="name" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta name="application-name" content="{{ .Title }} | {{ .Site.Title }}" />
+<title itemprop="name">{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}</title>
+<meta property="og:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
+<meta name="twitter:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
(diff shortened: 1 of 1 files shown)


gerbrent:
you're super quick @ChanceM! We really appreciate your contributions today. Top ~~notch~~ dynamic island.

gerbrent (approved):
tested and works as expected 👍🏻

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,9 @@
 <!-- Meta Ttitles -->
-<title itemprop="name">{{ .Title }} | {{ .Site.Title }}</title>
-<meta property="og:title" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta name="twitter:title" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta itemprop="name" content="{{ .Title }} | {{ .Site.Title }}" />
-<meta name="application-name" content="{{ .Title }} | {{ .Site.Title }}" />
+<title itemprop="name">{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}</title>
```
elreydetoda, line 2: Just a note for the future @ChanceM , another solution that's a bit more DRY for hugo is to use the [with function](https://gohugo.io/functions/with/).

While technically both accomplish what you're trying for, when I initially looked at this I was thinking "what is he trying to do...🤔". Because technically you're still accessing `.Title` even if the `.Title` if condition fails right now, because it's outside of the `if` statement. Hugo is inserting a blank space and HTML doesn't really care about that.

So, it's a larger [gulf of evaluation](https://en.wikipedia.org/wiki/Gulf_of_evaluation) (heard about this phrase recently and wanted to use it somewhere 😅) to understand why your solution works IMO.

If you'd used `with` you could've written it like this:
```suggestion
<title itemprop="name">{{ with .Title -}}{{- . }} | {{end}}{{ .Site.Title }}</title>
```

ChanceM: :+1:  Definitely still learning with Hugo, I appreciate the feedback!

PR #420, opened 2022-09-13
Add Matrix well-known client file

thefirstofthe300:
The Matrix spec has a defined process to allow clients to automatically discover the correct base URL for a home server using only the users MXID.

https://matrix.org/docs/spec/client_server/r0.4.0.html#well-known-uri

Adding this file will allow users to sign in by providing a client like element with `chrislas:jupiterbroadcasting.com` for example and the client will find the correct subdomain of the server.

### DIFF
--- a/static/.well-known/matrix/client
+++ b/static/.well-known/matrix/client
@@ -1,0 +1,5 @@
+{
+  "m.homeserver": {
+    "base_url": "https://colony.jupiterbroadcasting.com"
+  }
+}


noblepayne:
Thanks for the contribution @thefirstofthe300! 

For later, once #353 is merged, we could add an [e2e test](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/7f1a2aa7402f7cfdb250dce4ea5e3208c07420dc/test/e2e/test_infra_configs.py) for this.

elreydetoda:
> For later, once #353 is merged, we could add an e2e test for this.

Already added 😊

PR #423, opened 2022-09-15
Order slides based on show activity

### LINKED ISSUE #299: Show carousel - Show logos should rotate based on most recently aired show
image

The show whos episode aired most recently should be in position 1 and so on. Otherwise some shows never reach the front.

ChanceM:
Resolves #299 by ordering slides based on last episode of each show.

### DIFF
--- a/themes/jb/layouts/index.html
+++ b/themes/jb/layouts/index.html
@@ -19,36 +19,29 @@
 
     <div class="swiper show-slider">
       <div class="swiper-wrapper">
-
-       {{- range where .Site.Sections ".Title" "Shows" -}}
-        {{ $showScratch := newScratch }}
-          {{- range .Pages -}}
-            {{ $show := . }}
-            {{- range first 1 (where .Pages "Type" "episode") -}}
-              {{ $showScratch.SetInMap "episode-activity" (string (time .Date).Unix) $show }}
-            {{- end -}}
-
-          {{- end -}}
-          {{ $shows := $showScratch.GetSortedMapValues "episode-activity" }}
-            {{- range seq (sub (len $shows) 1) 0 -}}
-              {{- with index $shows . }}
-              <div class="swiper-slide">
-                <div class="card-image">
-                  <figure class="image">
-                    <a href="{{ print .Permalink }}">
-                      <img
-                      src="{{ print .Params.header_image | absURL }}"
-                      alt="{{ print .Title }}"
-                      class="show-img"
-                      />
-                    </a>
-                  </figure>
-                </div>
-              </div>
-            {{- end -}}
-            {{ $showScratch.Delete "episode-activity" }}
-          {{- end -}}
-        {{ end }}
+        {{- $all_most_recent := slice -}}
+        {{- range where .Site.Sections ".Title" "Shows" -}}
+         {{- range .Pages -}}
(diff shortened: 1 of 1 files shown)


elreydetoda:
First off, very nicely done figuring this out! When I saw @ironicbadger post that as an issue I thought...well that's a mighty ambitious request :sweat_smile: 

Now that you you did this PR though...you got my brain juices flowing :joy: 

---

So...this technically would always work because whenever JB wants to publish a new episode they have to push to production, so we can order shows by their most recent with Hugo instead of JS because it'll always have to rebuild on a new show! I didn't even think of that, I figured we'd have to do some weird JS to shuffle around the shows or something :sweat_smile: .

---

Since I saw this, you got me thinking about if there is a more "native" way to do this without the scratch stuff (which is definitely a valid option but is more complex IMO), and I think I figured out a way to do it (please let me know if you see any flaws/faults with my approach).

So, I pushed [a commit](https://github.com/elreydetoda/jupiterbroadcasting.com/commit/9439359450feeb66f5200cb6e075b7d982cb821d) to my repo's copy of your PR (wow that sounds convoluted... :sweat_smile: ), and instead of handling it via scratch and ordering it by a map of keys which are unix timestamps.

1. There's actually native [sorting mechanisms](https://gohugo.io/templates/lists/#order-content) with Hugo Pages objects (which is essentially a list page (if you do a `{{ print .Pages }}` it'll show you something like `Pages(6)` which means hugo treats it like a list Page with the same functions/variables)).
2. So, with that you can actually get the most recent Episode of each show (by default it give the oldest first, so you have to do `.Reverse` to get the most recent)
3. then make that into a slice of pages and then Hugo treats that slice as another Pages object (same as 1). (just found this one out by playing around, I didn't know that before trying to look at this :sweat_smile: )
4. So, you can take that Pages object and iterate over the most recent episodes, but instead of those episodes you can use the `.Parent` variable on each page, which'll point to the show's list page (the individual show pages (i.e. /show/self-hosted/))
5. Then it's like you're iterating over each show's list page and the variables don't even have to be changed.

I'll comment on the commit from my repo so it's a bit clearer, but I think I'd prefer the less complex route I was able to figure out (thank you for inspiring me @ChanceM :grin: ). So, if you don't mind modifying your PR to implement that, that'd be awesome.

I personally don't know what the `{{-` and `-}}` does (seems like removes whitespace from your commit message), but if you could keep that when you implement it (instead of my half-hazard attempt to figure out how to do it with all the extra whitespaces) that'd be awesome :grin: 

Again let me know if you have any questions/comment/concerns about my approach :upside_down_face:

ChanceM:
I will take a look at this tonight an update the PR. 
the {{- -}} removes the whitespace (which is ignored by the browser at presentation, but is still in the file being sent over the wire).

image

vs 

image

elreydetoda:
That is SOOO much nicer! 😅🤯🥳

I didn't even think to look at the HTML, but we probably should do that for all the hugo functions/go templating!

elreydetoda (approved):
LGTM, thanks for this PR @ChanceM!

gerbrent:
psst... this is so amazing you two, and a nice surprise for Chris and I to find on the home page this week - wonderful!!

PR #425, opened 2022-09-16
Only attempt setting src when livestream ID is found.

### LINKED ISSUE #328: PeerTube JavaScript included on every page
When I look at the browser console on any page other than the live page (e.g. https://new.jupiterbroadcasting.com/sponsors/), I get the following error:

```
(index):102 Uncaught (in promise) TypeError: Cannot set properties of null (setting 'src')
    at (index):102:73
```

It seems to be that the code embedded in the _jb-tube.html_ partial is included on every page, not just the live page.

https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/3216265083a3ea39b98950cc1443e49c4a981962/themes/jb/layouts/partials/live/jb-tube.html#L1

ChanceM:
Resolves #328 attempting to set src when the live stream id is not present.

### DIFF
--- a/themes/jb/layouts/partials/live/jb-tube.html
+++ b/themes/jb/layouts/partials/live/jb-tube.html
@@ -6,7 +6,12 @@
   <script type="text/javascript" src="{{ $jbTube.Permalink }}" integrity="{{ $jbTube.Data.Integrity }}"></script>
   <script>
     window.onload = () => {
-      jbLive().then(result => document.getElementById('liveStream').src = result);
+      jbLive().then(result =>  {
+        let liveStream = document.getElementById('liveStream');
+        if(liveStream)
+          liveStream.src = result
+        }
+      );
       doLiveHighlight();
     }
   </script>


CGBassPlayer (approved):
LGTM. Running locally removed the error from the console

elreydetoda (approved):
Based on @CGBassPlayer's approval (and my quick glance of the code (not that I don't trust @CGBassPlayer's approval, but I was personally curious how you did it :grin: )

For other's, he's querying the current page for an element with the [id of `liveStream`](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/0dfdc4f9b7bef542b078e52ebdd44ed65db6b912/themes/jb/layouts/partials/live/jb-tube.html#L17). If that comes back [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) (i.e. it exists) the he sets the `.src` attribute (which is the iframe).

He's not using a conventional formatting of an if statment (i.e. `if (condition) { action }`), but in JS you can implicitly have the action below if it's only one line. If it was more than one line he'd have to do the more conventional formatting.

PR #427, opened 2022-09-20
Add 404 page

### LINKED ISSUE #388: Create a 404 page
Currently there is no 404 page defined in the `404.html` and when a user hits a 404 they get an empty page with `404 page not found` text.

Current 404 page

Anyone have ideas for what the 404 page should look like?

CGBassPlayer:
Fixes #388 

* Add 404 page

I am not sure website nginx container is setup to automatically load the 404. But Hugo has a [small note](https://gohugo.io/templates/404/#automatic-loading)  that points to nginx's [docs](https://nginx.org/en/docs/http/ngx_http_core_module.html#error_page) about the error code handling


gerbrent:
@CGBassPlayer Looks good so far, any additional changes needed to make this functional as the 404 then? Wondering if this should be draft....? Ready for final review?

CGBassPlayer:
This should be set from the Hugo side of things. I think the configuration for the nginx should be checked to see if it is setup to automatically redirect to the 404 page when the site has a 404.

gerbrent:
^ Anyone willing to tackle this part?

CGBassPlayer:
I'm guessing we need to create an nginx config file to copy into the container? I can try it if everyone else is busy, but it would need someone who actually knows nginx to review it as I am pretty unfamiliar.

CGBassPlayer:
Converting back to draft for now. I will into the nginx config stuff in the meantime. If anyone has info on it, please let me know!

Dockerfile
```diff
@@ -5,4 +5,6 @@ RUN hugo
 
 FROM nginx:alpine
 RUN rm -rf /usr/share/nginx/html/*
-COPY --from=builder /site/public /usr/share/nginx/html
\ No newline at end of file
+COPY nginx.conf /etc/nginx/config.d/jb-404.conf
+RUN echo "\ninclude /etc/nginx/config.d/*.conf;" > /etc/nginx/nginx.conf
```
elreydetoda, line 9: Since we're leveraging the default config file in the nginx container ( default config uploaded [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/files/9651974/default.txt), typically a `.conf` extension but GH doesn't like that :sweat_smile: ), we can just un-comment the 404 line in the default config file.

```suggestion
RUN sed -i 's/#error_page/error_page/' /etc/nginx/conf.d/default.conf
```

Then we don't need the `nginx.conf` file at all as well.

### extra details

For those interested, I looked at the default container by shelling into the alpine container with this command: `docker container run --rm -it --entrypoint=sh --name jbcom nginx:alpine`. I copied it out of the container after starting it up (with the shell command) and then ran: `docker cp jbcom:/etc/nginx/conf.d/default.conf .`

Then I merged this PR (branch [here](https://github.com/elreydetoda/jupiterbroadcasting.com/tree/patch/modifying_e2e_GH_action-CGBassPlayer-feature/404-page)) into my E2E one (#353), because I added a way to easily build the prod container by running this: `docker build --build-arg='BASE_URL=TESTS_WEB' --rm -t jb-jbcom:latest '.'`. Which then let me view it locally with this: `docker container run --rm -it -p 1313:80 --name jbcom jb-jbcom:latest` (after I've already finished being in a shell for the other container).

elreydetoda (changes requested):
Everything else LGTM, after the changes are accepted and the `nginx.conf` is removed.

My only concern for the solution of us using nginx is if traefik will try and use it's own 404 page instead of allowing nginx to serve the 404 page. I'd imagine it'd default to nginx first (because it's essentially the "application"), but I don't personally know which one it'll use.

@ironicbadger, any thoughts on :arrow_up:?

.dockerignore
```diff
@@ -2,4 +2,5 @@ public/
 .github/
 .git/
 fireside-scraper
-scraped-data
\ No newline at end of file
+scraped-data
+nginx.conf
```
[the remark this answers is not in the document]
CGBassPlayer: I originally threw it in when I was thinking about it being built. I hadn't gotten to test to commit yet (moving changes to another machine) so thank you @elreydetoda for looking into this!

Dockerfile
CGBassPlayer: I appreciate the feedback on this @elreydetoda! I am not familiar with the deployment side of things (but I love learning about it) so this was more or less taking a guess at how to tackle this.

ironicbadger:
I would imagine that traefik will transparent in this process. I don't know for sure though.

Dockerfile
elreydetoda: No problem, you did awesome figuring out how to setup server blocks in nginx 🥳

Awesome! That is why I included a summary of how I figured it out, I'd hoped at least someone was curious 😅 (if not then it was at minimum a note to myself 🙃)

elreydetoda (approved):
Awesome! Well based on @ironicbadger 's testing in prod I think this LGTM. @CGBassPlayer mark it ready for review and I'll happily merge it 😁

CGBassPlayer:
Looks like it worked! Thanks everyone for the insight  😄 
404 in prod

Issue #432, opened 2022-09-27
Testing `develop` branch workflow


gerbrent:
Testing part of the workflow for our (soon) eventual move to a [Gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) workflow to protect main and offer - in the future - live full-production testing of the develop branch.

We'll see if I understand all of this well ; )

Steps for this test:
* [x] do a PR #431  to merge `main` -> `develop` to bring `develop` in sync w `main`
* [x] have someone merge said PR (thanks @elreydetoda !)
* [x] apply PR #400 to `develop`
* [x] see how the E2E Tests do w that one (for fun!). Not expecting anything PR #400 specific other than a green light..
* [x] If satisfied, do a PR from `develop` to `main` #433

gerbrent:
: )

![image](https://user-images.githubusercontent.com/7018928/192647442-846539f0-c637-4892-9c31-30babb6969fb.png)

gerbrent:
One gotcha:

The `develop` branch was auto-deleted on the merge to `main`,  since we have that set. 

![image](https://user-images.githubusercontent.com/7018928/192834184-fe51913b-7f82-4b9a-910c-ea32554593cc.png)

elreydetoda:
> One gotcha:
> 
> The `develop` branch was auto-deleted on the merge to `main`,  since we have that set. 
> 
> ![image](https://user-images.githubusercontent.com/7018928/192834184-fe51913b-7f82-4b9a-910c-ea32554593cc.png)
> 

Just a heads up, this could be prevented with branch protection rules. So while yes I did delete it by accident 😅, if we had branch protection rules on I wouldn't be able to. Only an admin might be able to, unless you required admins to adhere to the protection ( which I recommend you do enable this and unchecked that box when you need to force an action that's being blocked).

PR #400, opened 2022-09-07
Revamp Episode Layout

CGBassPlayer:
Fixes #226

* Shrink Hosts and Guests to a smaller image
* Implement tiles to allow vertical flow in content
* Entire Sponsor card is link instead of only the text

PR #433, opened 2022-09-27
`develop` -> `main` - testing w/ #400

gerbrent:
only https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/400 in this PR.

`develop` -> `main` to test workflow as discussed in #432


CGBassPlayer:
Woo! being the Guinea Pig!

elreydetoda:
@CGBassPlayer , everything still look good to you? (Just want to make sure)

elreydetoda:
Actually, looks like they're exactly the same (as expected :grin: )

image

Link you can use to look at it :grin: 

https://github.com/CGBassPlayer/jupiterbroadcasting.com/compare/bugfix/episode-layout...JupiterBroadcasting:jupiterbroadcasting.com:develop?expand=1

elreydetoda (approved):
LGTM based on this: https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/433#issuecomment-1260149902

elreydetoda:
I don't know why the branch was deleted...I didn't click the button, but I've restored it :sweat_smile:

gerbrent:
haha, guess we have to solve this one ; )
I imagine it'll be different when the default is set to `develop`, and no PR ever goes from `main`-> `develop`

image


Will work on that soon!

CGBassPlayer:
I know there are ways to protect a branch. All I know about it is that you can apply rules about branch (PR required, deployments need to be successful, etc). I found some [docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) about it, but I have never done this myself

CGBassPlayer:
So once I got home to the non-mobile version of GitHub, I managed to find it. If you go to the repo's settings -> Code and Automation -> Branches and you guys can setup branch rules and assign them to the appropriate branches. I might try to play with this on a test repo and report back what I think would be helpful

elreydetoda:
> haha, guess we have to solve this one ; ) I imagine it'll be different when the default is set to `develop`, and no PR ever goes from `main`-> `develop`
> 
> image
> 
> Will work on that soon!

So, it looks like your assumption was right, but now I can delete main :sweat_smile: 

image

You should be able to do something like this at bare minimum for develop

image

Then a no checks one for main, because this will prevent force pushes and deletions

image

This would be the ideal options (IMO) for develop, but we can ease into that if we want :upside_down_face: 

image

Ideally the "production" environment would be used for keeping your production secrets (i.e. ssh private key for connecting to the server) and it would require manual approval for deployment (similar to the infra repo's eventual setup), but that'd break how we currently use the scraper for auto-deployments.

At minimum though it should be set for protected branches for the deployment branches (after setting up the protected branches above).

Issue #244, opened 2022-08-08
Server Setup


StefanS-O:
Setup the server that JB provided for delivering the new site:

- [ ] finish Ansible scripts

List what's needed:

- rsync
- nginx (vhost for jupiterbroadcasting.net / archive. jupiterbroadcasting.net -> just static sites), will need more when we migrate Fireside
- letsencrypt
- user to deploy from  Github Action (Private Key needs to be set as a secret in Github Action, public key needs to be on the server)
- docker (for later stuff like search)

elreydetoda:
Oooh, if you want some help with this, I do this for work all the time 😁

Are you just doing a vhost with nginx @StefanS-O ?

We could commit the playbook to this repo to collaborate 🙃 

honestly I don't like doing 🔼 (committing the infra code beside the app code) because it can inhibit either the app or infra depending on if one doesn't iterate as fast as the other, but I think this is a nuanced case that might be an exception (plus I don't think anyone has enough rights to create like an IaC (Infrastructure as Code) repo or something like that 😅)

gerbrent:
@ironicbadger wants to sink his teeth into helping on the website, this might be the perfect place...

ironicbadger:
I will create an infra repo and migrate over the code from [here](https://github.com/selfhostedshow/infra) where it makes sense and then provide docs of how to use it.

Might be a day or two.

gerbrent:
@StefanS-O is also in the process of writing up some requirements to help that process. Go team!

StefanS-O:
@ironicbadger nice, i just added the required packages. Let's take a as little as possible approach and extend from there.

ironicbadger:
The site will be hosted on an existing Linode server that we use for several core jb services already.

I need to refactor a few things now that the purpose of the servers scope has increased somewhat over its original but fundamentals will remain the same as described [here](https://github.com/JupiterBroadcasting/infra/blob/auto_container_dns/ansible/group_vars/demo.yaml).

In order to proceed with deploying even the test site to this box I need to get a handle on how we intend to build the site into an artifact. I assume (and hope) this will be a container with hugo as discussed in #249 using Github Actions. We need to agree upon how to publish that artifact - is github container registry agreeable to all? Or should we clone this repo locally to the Linode box and build a local image there? I don't mind either solution but we need to make the decision.

elreydetoda:
> In order to proceed with deploying even the test site...

[...]

PR #436, opened 2022-09-29
Revert "testing rebase breakage"

elreydetoda:
Reverts JupiterBroadcasting/jupiterbroadcasting.com#435

sorry thought this was open against my personal repo... 🤦

### DIFF
--- a/test/e2e/test_home.py
+++ b/test/e2e/test_home.py
@@ -42,7 +42,6 @@ def test_dropdowns(page: Page, expected_dropdown_items):
         element = page.locator(selector)
         expect(element).to_contain_text(dropdown_item['title'])
     
-blah blah
 
 
 def test_nav(page: Page, expected_dropdowns, expect_nav_items):


gerbrent:
I believe this may have cause some deploy issue - git on the prod server was [complaining about diverging branches...](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/actions/runs/3155665115/jobs/5134572496#step:4:24)

We had to jump on the box, clear the local git repo and redeploy.. 
There can be methods for avoiding this in the future, since we do not really ever intend to need some git pushing functionality from the production box.

gerbrent:
Also: pushing to `main` should have more protections. This should, in best case, not be possible to have such an error break deployments... the consequences were small this time (no update of scrapes, therefore LAN missing from production) however, I wonder if it could potentially be more problematic.

elreydetoda:
> I believe this may have caused some deploy issue - git on the prod server was complaining about diverging branches...

That's so weird because it had 2 successful runs after that PR...

- https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/actions/runs/3151245985
- https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/actions/runs/3149008830


> We had to jump on the box, clear the local git repo and redeploy..

Just since you'd linked to it, the [`--ff-only`](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/actions/runs/3155665115/jobs/5134572496#step:4:30) `git config` option is the least interactive and most guaranteed to not screw things up.

> There can be methods for avoiding this in the future, since we do not really ever intend to need some git pushing functionality from the production box.

Also, another side note while I'm thinking about it, is if we'd gone with [an artifact approach](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/244#issuecomment-1220124933) to where that's the only thing the server is pulling down. It wouldn't care about these kinds of things. It just would pull down the container and keep running. It's because we're pulling down the repo on the prod server is why this git conflict error happened. Some could argue that's a feature, but I'm just giving facts about the situation 😅. my E2E PR already has the fundamentals if we wanted to use that approach since I'm already building the prod container and pushing to quay in it. Also, I'll happily jump on a call to help out get this setup if we want it.

> Also: pushing to main should have more protections.

Ya, I've outlined the branch protection we should use for main and 2 options for the develop [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/433#issuecomment-1261827926). Also @CGBassPlayer outlines where to go to add those branch protections.

gerbrent:
> That's so weird because it had 2 successful runs after that PR...

right?! I didn't know what to make of that.......

elreydetoda:
> > That's so weird because it had 2 successful runs after that PR...
> 
> right?! I didn't know what to make of that.......

Ya...I have no idea 😅

PR #443, opened 2022-10-04
Replace checkboxes with drop down in issue form

### LINKED ISSUE #390: Issue and/or PR Templates
Something I think would be useful for people users new to contributing to open source would be templates for creating issues and maybe PRs. I don't think it needs to be anything extravagant, but something to just let them know what they should put in  when submitting an issue and possibly add some auto tagging.

Initial idea I am thinking is:
* Bug Report
  * What is the bug
  * What device (Phone, tablet, desktop, etc.)
  * Steps to reproduce
  * Screenshots

* Feature Request
  * Describe the feature
  * Is it related to a problem?
  * Did you see this feature somewhere else?

* Bug Fix PR
  * Issue(s) fixed
  * What was done to fix it?

Looking for suggestions of templates or things in templates as well

CGBassPlayer:
Closes #390 
* Change checkboxes to dropdowns to prevent issue details being marked as tasks

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_shownotes.yml
+++ b/.github/ISSUE_TEMPLATE/bug_shownotes.yml
@@ -34,12 +34,3 @@ body:
       placeholder: |
         For dead links: If you find a working link please add it in! This can help us get the link updated faster.
         For typos, feel free to include the correction and a screenshot.
-  - type: checkboxes
-    attributes:
-      label: Contributing
-      description: Please read our [Contributing Guidelines](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/CONTRIBUTING.md) before submitting an issue.
-      options:
-        - label: I have read through the [Contributing Guidelines](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/CONTRIBUTING.md)
-          required: true
-#         - label: I have read and accept the [Code of Conduct](https://docs.github.com/articles/github-community-guidelines)
-#           required: true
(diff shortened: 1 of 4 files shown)


CGBassPlayer:
I may close this PR and open a new one to remove the commits from the scraper

gerbrent:
could try a force push on your repo's feature branch? would that work instead?

CGBassPlayer:
So when I look at the branch in my repo, it says I am only 1 commit ahead of develop instead of 4 so it may be fine when we merge this

gerbrent:
I agree, seems just fine now. Let us know when its ready for review!

gerbrent:
err, one consideration for dropdowns instead of checkboxes: it eliminates the possibility for the Issue author to make mulyipke selections, which I found useful already.

This feels like one of those conpromise scenarios.. no ideal path, just a choice of loss of functionality.

the Q becomes: do we prefer multi-selections or less issue-tracked checkboxes interpreted as "tasks" ? I'm not so sure....

Now I may br leaning toward keeping as is, cobsodering the above. Hmm!

What do you all think?

CGBassPlayer:
The drop down does allow multiple options to be selected if you enable `multiple: true` on the drop-down (which I did in this case to maintain functionality)

CGBassPlayer:
The drop down does allow multiple options to be selected if you enable `multiple: true` on the drop-down (which I did in this case to maintain functionality). This can actually be seen in the browser selection and device selection when someone reports a website issue.

CGBassPlayer:
We can talk about the contributing check box in the future

gerbrent:
Ok, I say drop the Contributor Guidelines check box - if someone actually wants to read it, they will. If not, they'll just check the box. Doesn't necessarily give us more or less information either way.

gerbrent:
Also: that IS a wonderful drop down feature - nice! Thanks for keeping me focused.

CGBassPlayer:
@gerbrent this is all set for your review

gerbrent (approved):

PR #445, opened 2022-10-04
local embed cdn file for pihole

### LINKED ISSUE #116: vjs.zencdn.net blocked by pihole
Just wanted to give a heads up that I use a [pi-hole](https://pi-hole.net) at home, and I'm pretty sure I use [this list of block lists](https://v.firebog.net/hosts/lists.php) (the Non-crossed lists group), and that domain (vjs.zencdn.net) was getting blocked (from [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/551e71a20d01b5fa9c8b87fd7d66dbc7613c116e/themes/jb/layouts/episode/single.html#L2)).

image

So, it might be worth considering just pulling that file into the project and serving it from JB's domain instead of the other domain. That way it won't get blocked, I'm just whitelisting it on my DNS for now but something to think about for other people.

devotoare:
Resolve #116

### DIFF
--- a/themes/jb/layouts/episode/single.html
+++ b/themes/jb/layouts/episode/single.html
@@ -1,7 +1,3 @@
-{{ define "headcontent" }}
-  <link href="https://vjs.zencdn.net/7.18.1/video-js.css" rel="stylesheet" />
-{{end}}
-
 {{ define "main" }}
   {{ if ne .Title (print .CurrentSection.Title " " .Params.episode) }}
     {{ .Scratch.Set "fullTitle" (print .Params.episode ": " .Title ) }}
--- a/themes/jb/layouts/partials/head.html
+++ b/themes/jb/layouts/partials/head.html
@@ -5,6 +5,7 @@
   {{ partial "meta.html" . }}
 
   {{ $sass := resources.Get "css/main.sass" }}
+  {{ $video := resources.GetRemote "https://vjs.zencdn.net/7.18.1/video-js.min.css" }}
   <!-- Defining options that production & development should have -->
   {{ $options := (dict "enableSourceMap" true ) }}
   {{ $style := "" }}
@@ -12,11 +13,11 @@
   {{ if hugo.IsProduction }}
     {{ $options = merge $options (dict "outputStyle" "compressed") }}
     {{ $style = $sass | toCSS $options | minify | fingerprint }}
-
   {{ else }}
     {{ $options = merge $options (dict "outputStyle" "expanded") }}
     {{ $style = $sass | toCSS $options }}
   {{ end }}
   <link rel="stylesheet" href="{{ $style.RelPermalink }}">
+  <link rel="stylesheet" href="{{ $video.RelPermalink }}">
 
 </head>


themes/jb/layouts/partials/head.html
```diff
@@ -5,18 +5,19 @@
   {{ partial "meta.html" . }}
 
   {{ $sass := resources.Get "css/main.sass" }}
+  {{ $video := resources.GetRemote "https://vjs.zencdn.net/7.18.1/video-js.min.css" }}
```
elreydetoda, line 8: I typically request people also `| minify | fingerprint` the files, so we can add some [security](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/a7585a36094491065cf88599b39c536eb68aad4e/themes/jb/layouts/partials/live/jb-tube.html#L6) (the `integrity` attribute) to the sources we're pulling in. Since this appears to already be minified (by the `.min` in the name), we shouldn't have to do the minification step.

```suggestion
  {{ $video := resources.GetRemote "https://vjs.zencdn.net/7.18.1/video-js.min.css" | fingerprint }}
```

themes/jb/layouts/partials/head.html
```diff
@@ -5,18 +5,19 @@
...
   {{ if hugo.IsProduction }}
     {{ $options = merge $options (dict "outputStyle" "compressed") }}
     {{ $style = $sass | toCSS $options | minify | fingerprint }}
-
   {{ else }}
     {{ $options = merge $options (dict "outputStyle" "expanded") }}
     {{ $style = $sass | toCSS $options }}
   {{ end }}
   <link rel="stylesheet" href="{{ $style.RelPermalink }}">
+  <link rel="stylesheet" href="{{ $video.RelPermalink }}">
```
elreydetoda, line 21: Adding the `integrity` attribute referenced [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/445/files#r990674593).

```suggestion
  <link rel="stylesheet" href="{{ $video.RelPermalink }}" integrity="{{ $video.Data.Integrity }}">
```

themes/jb/layouts/episode/single.html
```diff
@@ -1,7 +1,3 @@
-{{ define "headcontent" }}
-  <link href="https://vjs.zencdn.net/7.18.1/video-js.css" rel="stylesheet" />
-{{end}}
-
 {{ define "main" }}
```
elreydetoda, line 1: @devotoare, is there a reason why you moved the JS out of the single template?

I believe it was put there so it'd only load on that JS on the episode pages instead of every page.

You could still do the same thing you're doing with the `GetRemote` function w/Hugo, but just inside that template.

Based on my later feedback ([here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/445/files#r990674593) and [here](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/445/files#r990675234)), this'll integrate everything into this one file. (if we end up putting it back in this file, if not this can be rejected and the other ones applied)

```suggestion
{{ define "headcontent" }}
  {{ $video := resources.GetRemote "https://vjs.zencdn.net/7.18.1/video-js.min.css" | fingerprint }}
  <link rel="stylesheet" href="{{ $video.RelPermalink }}" integrity="{{ $video.Data.Integrity }}">
{{end}}

{{ define "main" }}
```

I'd typically also like to do a [check for production](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/a7585a36094491065cf88599b39c536eb68aad4e/themes/jb/layouts/partials/head.html#L12-L19), but since this is 3rd party code I don't think it's important to do that.

elreydetoda (changes requested):
First, I'd like to say thank you for doing a PR for this! :partying_face: 

I can't believe I didn't think about doing this! 🤦

I've got a [question for you](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/445/files#r990675579), and a few suggestions to help strengthen the security of us serving the content from JB's domain (overall it's not super important just a best practice IMO).

Overall though, everything looks pretty good! Thank you for doing this! :partying_face: :rocket:

gerbrent:
@devotoare - let us know if you want to go through the above suggestions. Just waiting on your input to move this one along..

And also: thanks for the PR!

gerbrent:
seeking extra clarification in #116 before taking action on this PR.

devotoare:
Sorry for not answering sooner, I rarely get time to do fun things. Looks like this is no longer needed as video is being removed per #116 (and also Chris on Office Hours though I don't know if that was an official decision). So I'm going to go ahead and close this out, and if we need it again for some reason or change our minds it can always be reopened with the changes implemented.

Issue #54, opened 2022-07-08
Sponsors - automate Sponsors page content from recent episode sponsors


gerbrent:
Curious if we could automate the creation/updating of text in the Sponsors page, pulling information on sponsors from /content automatically.

`current sponsors` = a recent time frame, something like "sponsors of a show published in the last 3 months"... need to make a decision there on specifics.

noblepayne:
Motivation is perhaps something like:
> "As a JB listener who is about to sign up for a company or service that is currently sponsoring JB, I would like to be able to find sponsor codes on the main website easily."

noblepayne:
Perhaps we need some notion of "active" (or perhaps the reverse, "expired") sponsors and campaigns? 

Alternatively perhaps just showing the sponsors from the most recent episode from each show would be a decent MVP version?

JR-git-hub:
Don't forget affiliate links that continue to be advantageous. They also remind past sponsors of the quality traffic JB drives.

So maybe a automatef highlighting of current sponsors but historical reference such that JB still gets affiliate credit related to back catalog plays, etc.

ChanceM:
I have a MVP of this that can either select from N number of recent episodes or N number of days back. Although we could remove the limit and list all sponsors. If we do limit by some factor I have it set to omit the show title if there are no sponsors found. Alternatively we could show a register interest link?

Jupiter.party is also listed as a sponsor in some episodes. I'm not sure if we would want to suppress that. 

![image](https://user-images.githubusercontent.com/11327907/189752066-276e53d9-f400-4ce6-a4a6-ddca09537998.png)

ChanceM:
@elreydetoda Care to take a look this [branch](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/compare/main...ChanceM:jupiterbroadcasting.com:feature/automate-sponsors-page) and see if you have any feedback? 

@gerbrent Probably need a JB decision for criteria for crawling / a method for designating active sponsors.

elreydetoda:
Sure, I'll try and look at it either tonight or sometime this weekend 😁

elreydetoda:
Dang...sorry @ChanceM I thought I'd remember to do this (going through GH notifications and stumbled across it again), do you mind just doing a draft PR? I look at the PR area a lot more than I remember looking through issues 😅

ChanceM:
[...]

PR #447, opened 2022-10-06
update sponsorship info

gerbrent:
In future, prefer method to be implemented by #54


gerbrent:
E2E tests totally caught an accidental typo that broke Hugo builds here - thanks @kylepotts ! 🎉 
(fixed w a quick commit and squash rebase)

CGBassPlayer (approved):

PR #452, opened 2022-10-08
Add Subscribe and Support on Show Page

### LINKED ISSUE #404: add Subscription (and Support) buttons to Show pages
Subscription and Support/Membership buttons, assuming #346 is merged, should be added to Show pages (i.e. https://www.jupiterbroadcasting.com/show/coder-radio/)

Proposed Subscribe button in #346 :
image

CGBassPlayer:
Fixes #404 

* On Show pages, add icons for subscribe (`bell-ringing-o`) and support (`usd`)
* On Episode Page, subscribe and support pages open in a new tab

### All shows have the new icons
LUP show page icons

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -34,7 +34,7 @@ paginate = 12
 
 # Main Menu
 [[Menu.Main]]
-    pre = '<img src="/images/logo_no_rocket.svg" style="height: 2rem">'
+    pre = '<img src="/images/logo_no_rocket.svg" alt="Jupiter Broadcasting logo without Rocket" style="height: 2rem">'
     identifier = "home"
     url = "/"
     weight =-100
--- a/content/show/coder-radio/0486.md
+++ b/content/show/coder-radio/0486.md
@@ -1,0 +1,79 @@
+{
+  "type": "episode",
+  "draft": false,
+  "show_slug": "coder-radio",
+  "show_name": "Coder Radio",
+  "episode": 486,
+  "episode_padded": "0486",
+  "episode_guid": "212084ca-b0c8-43b2-925e-59e423479fe4",
+  "slug": "486",
+  "title": "The Fight for the Next Knight Rider",
+  "description": "Why we think Google will ultimately lose the next big tech battle.",
+  "date": "2022-10-05T03:00:00-07:00",
+  "header_image": "/images/shows/coder-radio.png",
+  "categories": [
+    "Coder Radio"
+  ],
+  "tags": [
+    "andorid auto",
+    "android automotive",
+    "calyxos",
+    "carplay",
+    "coder radio",
+    "dev one",
+    "development podcast",
+    "flastseal",
+    "flatpak",
+    "grapheneos",
+    "infotainment",
+    "killed by google",
+    "onlive",
+    "oryx pro",
+    "permisions",
+    "qnx",
+    "sega channel",
+    "stadia",
+    "volvocar os",
+    "vscode"
+  ],
+  "hosts": [
+    "chris",
+    "michael"
+  ],
+  "guests": [],
+  "sponsors": [
+    "linode.com-cr",
+    "tailscale.com-cr",
+    "system76.com-cr"
+  ],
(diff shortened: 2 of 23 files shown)


elreydetoda:
So, I saw this PR and thought "there seems to be a lot of duplication for each shows with them all pointing to /membership and their corresponding show's subscribe page...there's got to be a more hugo native DRY way to do this... :thinking: "

It does require a bit of work, but I think in the long-term this would make things easier to maintain. I pushed up a branch on [my repo](https://github.com/CGBassPlayer/jupiterbroadcasting.com/compare/feature/subscribe-show...elreydetoda:jupiterbroadcasting.com:CGBassPlayer-feature/subscribe-show?expand=1) to kind of show as a reference for what I was thinking of. (because that would've been a lot of shifting around to explain and I wanted to make sure it worked before suggesting it)

---

Essentially it looks like you're just imitating the same behavior that was done before for the [`support_link` parameter](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/develop/content/show/office-hours/_index.md?plain=1#L15). While that does technically work it's a lot of duplicated effort since the support link points to the same place no matter what show it is.

---

In my branch I actually delete all those links, and then move the links outside of the [`with .Params.links` section](https://github.com/elreydetoda/jupiterbroadcasting.com/commit/106675976b9789eed7ef6e888420016d5248bf02#diff-ef32c4062b71c24d3627bd080d7eacaf081487a9ae260e7fcd00afb4930cf87eR24-R30), because IMO a show should always be able to point to the subscribe and support places no matter if there are other links for the show or not.

Then I use the [`.GetPage` hugo function](https://gohugo.io/functions/getpage/) to request the `/membership` page and get that page's hyperlink ([here](https://github.com/elreydetoda/jupiterbroadcasting.com/commit/106675976b9789eed7ef6e888420016d5248bf02#diff-ef32c4062b71c24d3627bd080d7eacaf081487a9ae260e7fcd00afb4930cf87eR28)), and I also get the relative `subscribe.md` file so when each show builds it's grabbing the one in it's folder.

I also removed the [support link's usage](https://github.com/elreydetoda/jupiterbroadcasting.com/commit/106675976b9789eed7ef6e888420016d5248bf02#diff-acc058a30ba83bdbb0115b13bb718d83988a5fea179961765fa907528d6e2d18R2) and corresponding params for each show.

---

Not saying you have to implement this, but I just figured I'd suggest it :grin: 

Awesome job working on this though!

CGBassPlayer:
I love the solution you came up this! I am not familiar with hugo yet (the functionality under the hood is a bit of a black box for me), so I just continued the convention that was already implemented. Thank you for making this PR so much better @elreydetoda

elreydetoda:
Awesome! I love the change :partying_face: 

I'll let @gerbrent give a final say for this since it's dealing with some more aesthetic stuff (and he was the original one you'd requested), but it LGTM :partying_face:

gerbrent:
As a narrow aesthetic opinion, this all looks good.
As a broader vision, I have opinions about these icons vs the buttons on Episode pages, which I prefer. But that's a larger conversation for a future issue and not appropriate in this PR.

Local testing:
* the buttons look great and work, 
* however they open as a new window/tab, which is in fact appropriate for every other icon in this section. Do we expect these two local links to open in the current tab/window, and the others to open in a new tab/window?

elreydetoda:
> Do we expect these two local links to open in the current tab/window, and the others to open in a new tab/window?

I personally think all things that keep you on the JB website shouldn't open a new tab, but that just my 2 cents :upside_down_face:

CGBassPlayer:
>  Do we expect these two local links to open in the current tab/window, and the others to open in a new tab/window?

I was just keeping the behavior consistent across all the icons. I felt like if a user had used the other icons and they open in a new tab, and then 2 of them don't, would they think it is a bug?

gerbrent:
i say implement as is and lets have this discussion as a larger review of the Show pages in the future.

PR #456, opened 2022-10-09
Deploy Previews and URL Encoding Tags

### LINKED ISSUE #257: deploy previews
### Discussed in https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/discussions/182

<div type='discussions-op-text'>

<sup>Originally posted by **elreydetoda** July 24, 2022</sup>
So, one thing that I've normally done in the past is hooked up Netlify to the repo. Normally this is for easy of hosting the NetlifyCMS, but a passive added benefit is using them for deploy previews. This means it'll automatically deploy the changes proposed in a PR for the website. It can be used on the free tier and if layer down the road it breaks or cost money. It'll be easy to remove it, and we'll go back to doing what we're currently doing.

So, this is an example past website that has a deploy preview: https://github.com/Climate-Refugee-Stories/crs-website/pull/95

Then this is the config that's used: https://github.com/Climate-Refugee-Stories/crs-website/blob/master/netlify.toml

It'll take someone on the JB side to hook it up (I'd be willing to hop on a call and work through it with them), but it'd be nice to see peoples change *and* interact with them instead of relying on people sending screenshots. 😅

There are more advanced features which you'd have to pay for, i.e. a utility overplayed on the website for screenshot'ing + commenting while you're navigating the deploy preview, but just the actual deploy preview feature is free.

We could eventually hook everything up to do deploy previews based on where the changes were coming from but till then just the JB.net works.</div>

### LINKED ISSUE #319: Non-safe URL linking from tags
While attempting to do #257 

I found out we have `#` signs throughout our tags, and this isn't being handled properly in  hugo's output as well.

On this episode: https://new.jupiterbroadcasting.com/show/coder-radio/428/

They reference `c#` the programming language. I found out it's actually linking to the `tags/c` because it's thinking the `#` is used for an html reference to an id (or something like that), but it isn't retaining it.

image

elreydetoda:
closes #319 
closes #257 

While we're not going to be able to use netlify, I found render.com on the [hugo deployments/hosting website](https://gohugo.io/hosting-and-deployment/) (here is [their hosting page](https://gohugo.io/hosting-and-deployment/hosting-on-render/)).

Render doesn't have any issues with special characters for file names (like [netlify did](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/257#issuecomment-1220221546)), so we won't have to modify any of the content and this PR also fixes #319 as well.

I'll happily hop on a call to help walk anyone through this that's going to be setting this up (if wanted, and we could probably do it along side of #353).

---

### The downsides to render.com (similar to netlify)

- there's a limited amount of deploy hours we'll get being on the free plan: https://render.com/docs/build-limits
  - a simlar limitation we'd have on netlify, but is expected since we're on a free plan
  - if we run out of build minutes (unless JB wants to pay for more), then we go back to the [manual way](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/257#issuecomment-1227824750)
  - here is their SSG (static site generate) information: https://render.com/docs/static-sites
- we'll never actually be deploying the main branch
  - this is fine and not needed since we're using the git-flow workflow #432, and our main is being deployed to Linode
- (netlify wasn't like this) It takes ~14 minutes to deploy the whole site...
  - while this sucks, it's better than not having any deploy previews. Plus this is supplementary for PR reviews to easily see what the changes would do not for people developing the changes.
    - all changes should be checked locally by the person who's submitting the PR

---

### Misc

I've added a render.yaml file to the repo as well, so this'll allow us (once JB signs up for a render account and authorizes render access to this repo (I'll be happy to hop on a call and work through this)) to easily deploy the exact configuration I have setup on my repo for JB's render account. Docs on that file are here:
- https://render.com/docs/infrastructure-as-code
- https://render.com/docs/blueprint-spec

---

### What it'll look like

This is what it'll looks like to choose the main branch and setting the name for the project on render
image

This is what the deployment will look like (you'll go look at under the <project-name> -> Events -> Most recent deploy (click the `Deploy` word))
image

This is what it'll look like when the deployment will finish
image

This is an example of a PR build: https://github.com/elreydetoda/jupiterbroadcasting.com/pull/19#issuecomment-1272554896
(no one can see the status unless they're added to the project, but everyone can see the deploy preview after it's deployed)

---

### reasoning for why to not use netlify currently

I'd honestly prefer netlify, but netlify [not handling the c# tag file](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/257#issuecomment-1220221546)...is pretty much absolutely ridiculous...we shouldn't have to change our content so we can use deploy previews 😞

Maybe we can eventually migrate to using netlify, but it's going to be a workflow change for Chris and the whole JB team (which is the opposite of what we wanted for this migration). we shouldn't have to make a lot of modifications for a non-critical needs (since deploy previews are only supplementary) and hugo is actually working with it currently so it's only a limitation on netlify's side...

Maybe whenever the main website becomes the source of truth for the shows we could transparently handle adding c# tags as c-Sharp or something like that, but till then I think it's stupid that netlify doesn't support that 🤦 and we shouldn't have to do a lot of work to make their product work. (IMO)


gerbrent:
XKCD reference makes me like them already!

image

gerbrent:
merging this in real time together w @elreydetoda - woot!

elreydetoda:
Cool feature about the deploy previews, that I didn't know about till it was merged, was that it actually creates a deployment environment on the repo.

image


Which brings you to this [page](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/deployments/activity_log?environment=develop+-+jb-deploy-prv):

image

And if you click on the "Deployed" hyperlink will take you to the deployed site

PR #74, opened 2022-07-10
#67 - Added Conditional in Preview Title

Lord0fBytes:
From my comment on Issue #67 
https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/67#issuecomment-1179721718

PR #463, opened 2022-10-16
Add the show name to the episode page titles

### LINKED ISSUE #459: Episode - browser page title to include show name
### Feature Description

On Episode pages, the browser tab/window title does not include the Show name, only the Episode name in the form `Episode Name | Jupiter Broadcasting`

Would be more desirable to have the Show name and Episode number included aswell, such as:

`Episode Name` | `Show Name` `Episode number` | `Jupiter Broadcasting`

^ is the same formatting found on Episode cards throughout the website.
Note LAN likely is an exception to this titling, as was seen in #67 

---

### Current Show window title
image

### Current Episode window title
image

### Current Episode Cards (reference for title formatting)
image

### Reference to Feature

_No response_

### Contributing

- [X] I have read through the [Contributing Guidelines](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/CONTRIBUTING.md)

komish:
Fixes #459

As I understood it, this change required changing the page meta's Title properties, as well as any related properties. I understood "related properties" here to mean, effectively, any properties that had the same template syntax within the `meta.html` partial.

To address LAN, I copied the same conditional applied in #74.

Finally, I had to change the `" "` used in the `print` function call to use backticks to avoid a collision with some of the template properties that were using double quotes.

Thanks for participating in Hacktoberfest!

Signed-off-by: Jose R. Gonzalez <jose@flutes.dev>

### DIFF
--- a/themes/jb/layouts/partials/meta.html
+++ b/themes/jb/layouts/partials/meta.html
@@ -1,9 +1,22 @@
-<!-- Meta Ttitles -->
-<title itemprop="name">{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}</title>
-<meta property="og:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta name="twitter:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta itemprop="name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta name="application-name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
+<!-- Meta Titles -->
+{{- $titleMeta := .Site.Title -}}
+{{- if .Title -}}
+    {{ if and 
+        (and (isset .Params `show_name`) (isset .Params `episode`)) 
+        (ne .Title (print .CurrentSection.Title ` ` .Params.episode))
+    -}}
+        {{/* for all other shows */ -}}
+        {{- $titleMeta = (printf `%s | %s %v | %s` .Params.show_name .Title .Params.episode .Site.Title) -}}
+    {{ else -}}
+        {{/* covers LAN and any page that does not have a show name or episode */ -}}
+        {{- $titleMeta = (printf `%s | %s` .Title .Site.Title) -}}
+    {{ end -}}
+{{ end -}}
+<title itemprop="name">{{ $titleMeta }}</title>
+<meta property="og:title" content="{{ $titleMeta }}" />
+<meta name="twitter:title" content="{{ $titleMeta }}" />
+<meta itemprop="name" content="{{ $titleMeta }}" />
+<meta name="application-name" content="{{ $titleMeta }}" />
 <meta property="og:site_name" content="{{ .Site.Title }}" />
 
 <!-- Meta Description -->
--- a/themes/jb/layouts/partials/meta.html
+++ b/themes/jb/layouts/partials/meta.html
@@ -6,7 +6,7 @@
         (ne .Title (print .CurrentSection.Title ` ` .Params.episode))
     -}}
         {{/* for all other shows */ -}}
-        {{- $titleMeta = (printf `%s | %s %v | %s` .Params.show_name .Title .Params.episode .Site.Title) -}}
+        {{- $titleMeta = (printf `%s | %s %v | %s` .Title .Params.show_name .Params.episode .Site.Title) -}}
     {{ else -}}
         {{/* covers LAN and any page that does not have a show name or episode */ -}}
         {{- $titleMeta = (printf `%s | %s` .Title .Site.Title) -}}


komish:
I just realized that this PR targets the `develop` branch and I'm not sure if that's the goal. Please let me know if I needed to instead target `main` or something else.

CGBassPlayer:
Pointing at the `develop` branch is correct. We do this to protect the `main` branch which is what generates the actual website.

I will review THIS PR tomorrow. Thank you for working on this!

komish:
I just noticed https://github.com/JupiterBroadcasting/jb-hugo-theme/blob/main/layouts/partials/meta.html - should this change have been made there instead?

elreydetoda:
> I just noticed https://github.com/JupiterBroadcasting/jb-hugo-theme/blob/main/layouts/partials/meta.html - should this change have been made there instead?

Not yet, we haven't done the separation of the theme that we'd created that for.

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,9 @@
 <!-- Meta Ttitles -->
-<title itemprop="name">{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}</title>
-<meta property="og:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta name="twitter:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta itemprop="name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
-<meta name="application-name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
+<title itemprop="name">{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (print .CurrentSection.Title ` ` .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end }}{{ .Site.Title }}</title>
```
CGBassPlayer, line 2: So this is more preference than anything, but I think leaving these Hugo blocks in a long one-liner is *very* difficult to read. I would love to see it broken up into separate lines.

komish: I can give it a shot this evening! The challenge I had was making sure I didn’t inadvertently linebreak a field once rendered. I’ll play with the template linebreak handler logic and report back.

CGBassPlayer: I could be wrong, but I thought it was similar to a `p` tag where it would not insert a newline unless you added a line break tag (`<br/>`) or code ( I think either  `&#013;` or `&#010;`? Don't quote me there).

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,9 @@
...
+<meta property="og:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (print .CurrentSection.Title ` ` .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end}}{{ .Site.Title }}" />
```
elreydetoda, line 3: Also, something that might make your `if` a bit clean for your `print` is to use [`printf`](https://gohugo.io/functions/printf/) instead. With something like this:
```suggestion
<meta property="og:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (printf `%s %v` .CurrentSection.Title .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end}}{{ .Site.Title }}" />
```

elreydetoda (commented):
Just had some thoughts and figured I'd jot them down since I was already here responding to the HTML comment (and had a bit more time today :sweat_smile:)

themes/jb/layouts/partials/meta.html
komish: @elreydetoda @CGBassPlayer 

Roger that. I think I understand that what's rendered won't linebreak unless we have an explicit linebreak tag. For the `<title>` tag, I imagine it shouldn't be a functional problem; it would just look funny to have the source broken up. In the `<meta>` tags that follow, though, I wasn't as sure because the value we're modifying are in attributes instead of contained between tags. I don't do a lot of web dev, so I'm not sure how that plays out.

All that said, I think I've got this reorganized in a succinct way. PTAL when you've got a moment!

themes/jb/layouts/partials/meta.html
komish: @elreydetoda Regarding using `printf`, the only reason I used this block (as is) is because it's used elsewhere for exactly this condition (i.e. for matching LAN and avoiding duplication). To that end, I could switch this, but it would mean you'd have two conditional evaluations in your code that are intended to produce the result.

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,9 @@
...
+<meta name="twitter:title" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (print .CurrentSection.Title ` ` .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end}}{{ .Site.Title }}" />
+<meta itemprop="name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (print .CurrentSection.Title ` ` .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end}}{{ .Site.Title }}" />
+<meta name="application-name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ if ne .Title (print .CurrentSection.Title ` ` .Params.episode) }}{{ .Params.show_name -}}{{- if .Params.show_name }} {{ if .Params.episode }}{{ .Params.episode -}}{{end}} | {{end}}{{end}}{{ .Site.Title }}" />
```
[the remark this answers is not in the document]
komish: Sure thing. The reason I avoid `with` here was simply because the scoping change caused it read a bit strange to me. That is, `{{ with .Title }}` assigns `.Title` to `.`, and so it read strange to me that `{{ . }}` only contains the contents of `.Title` (here, a string), where everywhere else, I expect `{{ . }}` to mean some _object_.  That said, that's just me - I can make this change. No problem at all.

I'll adjust/accept this and repush/squash.

themes/jb/layouts/partials/meta.html
komish: For clarity, I copied it from #74.

themes/jb/layouts/partials/meta.html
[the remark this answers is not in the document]
komish: Actually, I'm not sure this still applies with the revised logic that I used in order to make it a bit more organized/clear in response to the previous comment from @CGBassPlayer. PTAL and let me know if using `with` here still makes sense.

komish:
@elreydetoda I went ahead and used custom variables which cleans things up a bit more.

CGBassPlayer (approved):
LGTM

gerbrent:
Nice work everyone!!

This is functionally working, but a little formatting is required:

The original [issue](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/459) suggested the following format:
`Episode Name` | `Show Name` `Episode number` | `Jupiter Broadcasting`

However, the current implementation generated the following:
`Show Name` |  `Episode Name` `Episode number` | `Jupiter Broadcasting`
"Self-Hosted | The Badger Stack 81 | Jupiter Broadcasting"

image


i.e. should be:
"The Badger Stack | Self-Hosted 81 | Jupiter Broadcasting"

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,22 @@
...
-<meta name="application-name" content="{{ .Title -}}{{- if .Title }} | {{end}}{{ .Site.Title }}" />
+<!-- Meta Titles -->
+{{- $titleMeta := .Site.Title -}}
+{{- if .Title -}}
+    {{ if and 
+        (and (isset .Params `show_name`) (isset .Params `episode`)) 
+        (ne .Title (print .CurrentSection.Title ` ` .Params.episode))
+    -}}
+        {{/* for all other shows */ -}}
+        {{- $titleMeta = (printf `%s | %s %v | %s` .Params.show_name .Title .Params.episode .Site.Title) -}}
```
gerbrent, line 9:
```suggestion
        {{- $titleMeta = (printf `%s | %s %v | %s` .Title .Params.show_name .Params.episode .Site.Title) -}}
```

themes/jb/layouts/partials/meta.html
```diff
@@ -1,9 +1,22 @@
...
+    {{ else -}}
+        {{/* covers LAN and any page that does not have a show name or episode */ -}}
```
gerbrent, line 11: LAN looks good!

gerbrent (changes requested):

gerbrent:
Also happy to push this w above changes for `hacktoberfest` before the end-of-month, even if @elreydetoda is away on a fun holiday adventure this week.

gerbrent:
Also thanks for the contribution @komish ! Happy to have another Red Hatter among us ; )

komish:
@gerbrent good catch! I’ve got limited access to computers at the moment so I’ve committed the suggested change as you provided it (from mobile).

Thanks for participating in Hacktoberfest!

gerbrent:
Tested again and looks good! 👍🏻 🎉

gerbrent (approved):

elreydetoda:
Just wanted to comment and say awesome job @komish! It looks really slick now :grin:

PR #464, opened 2022-10-17
This increases the heading size of topic in footer

### LINKED ISSUE #245: footer - titles could be a bit bigger.
Today is my first visit to the new jupiter broadcasting site. It's looking awesome!

I noticed I found myself clicking the column labels instead of the various items in the footer. I think "Get Social" , "Center", "Right" should be a slightly bigger font. I know the column labels are bold but I thought that meant I was on that page already. I also think if  you aren't going to use the center and right columns they just shouldn't be visible. 

image

BenjaminHCCarr:
- This should increase the size of the topics in the JB footer
- This should close #245

### DIFF
--- a/themes/jb/layouts/partials/footer.html
+++ b/themes/jb/layouts/partials/footer.html
@@ -8,7 +8,7 @@
             </a>
           </div>
           <div class="column has-text-centered-mobile">
-            <h6 class="title is-6">{{ .Site.Params.footer.left.headline }}</h6>
+            <h4 class="title is-6">{{ .Site.Params.footer.left.headline }}</h4>
             <ul>
               {{ range where site.RegularPages "Section" "community" }}
                   {{ $link := "" }}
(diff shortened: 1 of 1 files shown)


CGBassPlayer (changes requested):
So in my local testing, changing the tag to `h4` does not change the size of the text for me

### PR
pr-footer

### Current Website
website-footer

What you may need to do is add the `is-size-5` class to the tag in order to get the text larger. Here is a link to the [documentation](https://bulma.io/documentation/helpers/typography-helpers/#size) about it.

CGBassPlayer (approved):
Much better!

image

I also appreciate the removal of the `is-6` class as well since it did not make much sense having there in the first place... But this LGTM

elreydetoda:
Sounds good!

Merging, based on @CGBassPlayer's approval, to develop so we can push to the website whenever we deploy next :grin: 

Thanks @BenjaminHCCarr for the contribution! :partying_face: :rocket:

timlinux:
@elreydetoda not good to merge with failing tests though?

CGBassPlayer:
So that e2e test fails when you do not have write permissions to the repo. It's failing trying to generate the report and I'm not sure we have a way around that at this point.

elreydetoda:
Yep, @CGBassPlayer is right. I've got a temporary patch for it in #353 (at the bottom of the "GH actions" section in [this comment](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/353#issuecomment-1240682632)).

PR #467, opened 2022-10-22
Improvements to makefile: help system and minor tweaks

timlinux:
If you are using VSCode then it is nice to be able to run the make tasks from within the IDE nicely:

image

You can access them from the Run and Debug Tab (1) and then the Run tool (2).

The PR also adds some niceties to the Makefile:

1. Adds a help target  ``make help``
2. Adds a help string for each target

Example usage:

image

Three last things to note: 

1) I believe the ``dev-build-metrics`` target is inconsistently named - I have renamed it ``dev-metrics`` to be more consistent with the ``dev`` target which it emulates.

image

2) Maybe I missed it but the makefile references a docker-compose file for the make ``deploy-prod`` target which I cannot find anywhere in the source tree.

image

3) The docker image build names are inconsistent - some with hyphens separating others with underscores. It would be nice to harmonise these to mimimise cognative friction :-P


CGBassPlayer:
I guess now is the time to decide if we want IDE configurations in the project. Looking for feedback on this one...

CGBassPlayer:
> 2. Maybe I missed it but the makefile references a docker-compose file for the make deploy-prod target which I cannot find anywhere in the source tree.

So I could be wrong but I believe that the linode that hosts the website contains that docker-compose file. It contains the configuration for deploying the container while the everything that is in the repo focuses on building the image itself.

timlinux:
> First of all, amazing job on this! Love the detail in the PR. But I think we need to decide if we want IDE configurations in the project. Looking for feedback on this one...

Aww, thanks. For background, I am busy working on a revamp of our website for https://qgis.org (currently in sphinx) and I am planning / probably going to use Hugo for it. Since I listen to the jb podcasts I thought I would base our website loosely on the work you guys are doing and upstream any tweaks and improvements I can to you. Feel free to reject anything that is not interesting, I won't be offended. If you want I can strip out the VSCode piece and just include the Makefile improvement.

gerbrent:
Thanks for sharing that background Tim and agree - this is all great!

It would indeed be helpful to have the makefile improvements stay here (ready to merge soon!), w the VSCode change in a new PR since it may linger a little longer..

But/and both great contributions! Can't wait to see what else comes our way thanks to your efforts... and let us know if you need specific help w QGIS !

Makefile
```diff
@@ -1,17 +1,28 @@
...
+		@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort  | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m - %s\n", $$1, $$2}'
+ 
+dev-build: ## Generate the development docker container
 	docker build --rm -f Dockerfile.dev -t jb_hugo:latest .
 
-dev: dev-build
+dev: dev-build ## Run the development hugo container
 	docker run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest
 
-dev-build-metrics: dev-build
```
gerbrent, line 7: really appreciate this change - good one!

Makefile
```diff
@@ -1,17 +1,28 @@
...
+dev-metrics: dev-build ## Run the development hugo container with metrics
 	docker run --rm -it -v "$${PWD}":/app -w /app -p $${HOST_IP:-127.0.0.1}:1313:1313 jb_hugo:latest serve -D --bind 0.0.0.0 --templateMetrics --templateMetricsHints
 
-deploy-prod:
+deploy-prod: ## Deploy everything to production
 	docker build -t jb-jbcom .
 	docker-compose -f ~/docker-compose.yml up -d jb-jbcom
```
gerbrent, line 23:
```suggestion
	docker-compose -f ~/docker-compose.yml up -d jb-jbcom 
	# docker-compose.yml located on the production host, not within this repo.
```

Simply adding the answer to your questions "where is the compose file?" for others w the same question in the future.

gerbrent (changes requested):
Approving the makefile contributions only, not the VSCode contributions (to be reviewed in the future, in a separate PR)

all looks good when tested locally, and love the functionality and thoughtful make command descriptions. Thanks @timlinux !

timlinux:
@CGBassPlayer @gerbrent Thanks for the good feedback - I have stripped out the vscode related bits as requested.

CGBassPlayer:
I will review the changes one more time as soon as I can

CGBassPlayer (commented):
Everything LGTM once @gerbrent's [comment suggestion](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/pull/467/files#r1008194143) is is added and this should be good to merge.

The `dev-metrics` command change is the only thing that could be considered a "breaking change" to contributing users' configurations and personal scripts. We may want to throw a message in the Matrix chat when this is merged just in case. But probably not as big as an issue as I am making it out to be as well.

gerbrent:
thanks Tim!

Would also be great if you could submit the VSCode features to a new PR since you already did all the work - hopefully that's still trivial for you at this point.

gerbrent:
✔️ Matrix chat notified about the potential breaking change.

timlinux:
> 

@gerbrent sorry yes it was clear for me - the VSCode bit is on its way in a separate PR 🚀

Makefile
```diff
@@ -1,17 +1,29 @@
...
+help:
+		@echo "Jupiter Broadcasting Hugo Project"
+		@echo 
+		@echo "Help for using this Makefile"
+		@echo
+		@echo "For detailed help about this project please visit:"
+		@echo "https://github.com/JupiterBroadcasting/jupiterbroadcasting.com"
+		@echo
+		@echo "------------------------------------------------------------------"
+		@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort  | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m - %s\n", $$1, $$2}'
```
elreydetoda, line 10: Reviewing #478 and just wanted to identify what you're doing on this, personally and for others. (going to explain what I understand so far and @timlinux could you explain what I don't or I get wrong? :upside_down_face: 

So the grep command (`grep -E`) is using [extended regex](https://www.computernetworkingnotes.com/linux-tutorials/use-extended-regular-expressions-with-grep-command.html), which is more portable than PCRE since it's POSIX compliant (IIRC).

regex explained:
* This regex `^[a-zA-Z_-]+` is looking for lines that start with any alpha character + `-` & `_` and the `+` says one or more consecutive characters. 
* then there's the literal character of `:`
* after that it's a little loose with `.*?` and can actually be replaced with something like: `[a-zA-Z_ -]+`, but either way it works
* this matches up till the `##` which is the comments right after each make target, which this means all make targets in the future will require a comment to be included in the help section
* ` .*$$` - this then match a literal space character and anything after that (the comment) till the end, which requires 2 `$`, so make doesn't interpret the `$` as something else.

The `$(MAKEFILE_LIST)` appears to list the possible make files (which if we had more should grab all files), but for now it just lists `Makefile`

The sort just sorts the make targets (which sort should just do alphanumeric sort), but could be forced with `-d` which is a dictionary sort (alphanumeric).

Then awk is being used to print the output from the grep command.
* `BEGIN` - I'm not super familiar with `awk`, so I don't know what this does
* Next is the field separator (`FS`) begin defined (or that's what google says :upside_down_face:), which means all the content between the actual commands and the `##` are used to essentially split the content as a delimiter
* the printf command is how the content is getting printed out
* the `\033`, `[36m`, and `[0m` all appear to be color escape things (don't exactly know how it works, but I've seen it in scripts before
* the `%-20s` is essentially saying print a string (typically `%s`) and it's "width" is 20 characters, but fill it with spaces if there's not that many characters
* then it's the literal characters of ` - `
* follows by the next string: `%s`
* followed by a newline character `\n`
* the `, $$1, $$2` says it should print the first string and the second string respectively to their positions, and the `$$` is used for ensuring make doesn't interpret those `$` characters.

PR #479, opened 2022-10-29
Added runners for vscode

timlinux:
If you are using VSCode then it is nice to be able to run the make tasks from within the IDE nicely:

image

You can access them from the Run and Debug Tab (1) and then the Run tool (2).

See #467 for preceding discussion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ $RECYCLE.BIN/
 
 # Editor
 .idea
-.vscode
 .kateproject
 .kateproject.local
 .fleet
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,50 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Dev Build",
+            "type": "shell",
+            // see tasks.json for definition
+            "preLaunchTask": "dev build", // The name of the task defined above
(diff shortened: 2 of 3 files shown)


CGBassPlayer:
I plan to look into this soon. The main things I’m looking for is
* Will this config mess with peoples current project configuration?
* Does this configuration work on Windows and Linux?
* Will others adding custom run configurations (like using Hugo on host instead of docker) have a high likelihood on sneaking into a commit and changing others’ configs?

—-

While I am looking at these, how do other people feel about keeping run configuration for editors and IDEs in the repo?

If we decide yes, I can add my current configuration for [Fleet](https://www.jetbrains.com/fleet/) as well once it is out of public preview (still kind of buggy I’m it’s current state). I think those 2 and maybe a stripped down `.idea` folder for more serious IDEs (PyCharm id the one that comes to mind.

elreydetoda:
So, personally I think this could lead down a very long rabbit hole :sweat_smile:.

A good compromise (IMO) would be to have something like a `.ide` folder and then we can put all the custom configurations there (inside their respective folders). I think keeping those configurations lean (like @CGBassPlayer player had mentioned with the `.idea` folder) would be a requirement. This would allow people to just copy the folder to the expected location easily and not allow the:

> high likelihood on sneaking into a commit and changing others’ configs

to happen.

gerbrent:
great suggestion! I say lets implement this path, and update the docs to reflect this option/feature/asset is available too

CGBassPlayer (changes requested):
I agree with @elreydetoda's suggestion of creating a `.ide` folder with the `.vscode` folder within it. That will allow new contributors to have the default commands, but not worry about writing over others IDE configurations.

PR #484, opened 2022-11-09
Fix Periodic Test

CGBassPlayer:
So I just looked at the periodic tests and noticed that they were failing do to a typo in Job Trigger. This should fix that

### DIFF
--- a/.github/workflows/periodic_test.yml
+++ b/.github/workflows/periodic_test.yml
@@ -2,7 +2,7 @@ name: Periodic Tests
 
 on:
   schedule:
-      - 0 12 * * 0 # Noon on Sunday UTC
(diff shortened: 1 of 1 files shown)


elreydetoda (approved):
weird...we might need a way to test this kind of stuff on submission so we don't have something like this happen again. Doesn't have to run them, but just some type of validation tool :thinking:

CGBassPlayer:
I agree. The changes in e2e are automatic sine the PR triggers them so I didn't even think about this at first. A validation tool would be good. I wonder if using act would be of use to us.

elreydetoda:
ya, [act](https://github.com/nektos/act) probably would be a good idea. Just have that run as a validation step for PRs that modify things in `.github/workflows/*`

PR #499, opened 2023-02-04
Adding standard favicon and svg favicon for theme support

### LINKED ISSUE #198: No favicon on jupiterbroadcasting.com
Currently, jupiterbroadcasting.net does not have a favicon, which means the browser displays the default icon in the browser tab:
image

For comparison, this is what jupiterbroadcasting.com currently looks like:
image

TheHesster:
This is my proposed solution to the JB favicon.

I have added an SVG favicon to enable browser theme to switch between a white JB rocket and a #2a2a2a JB rocket as seen here.

JupiterDarkOnLight
JupiterLightOnDark

This isn't supported by every browser or even many places in current browsers like bookmarks and recent pages it seems.

So I have also updated the standard favicon to be a #2a2a2a background with white JB rocket as seen here.

JupiterDefault

closes #198


gerbrent:
Here's what I'm seeing:
(note bookmarks too)

# Firefox
image

# Chromium
image

TheHesster:
Yep that's what I was seeing too. No matter what theme I added to Chrome it wouldn't change. All documentation about svg favicons says that Chrome has this support. Maybe not Chrome for Linux if there is a difference? 
I haven't tested Chrome on a Windows box connected to the site to see if that works though...

gerbrent:
One curiosity in specific: 

See the dimension skewing of the Firefox bookmark favicon? A notable fat-but-short rocket?
Any thoughts on how to avoid this?

The chromium rocket seem squished too in comparison to the Firefox URL favicon, and to the original [source rocket](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/develop/static/images/logo.png).

TheHesster:
I have a feeling your Firefox bookmark icon is an old cached one. The only way for me to get mine to change was to delete all history (all boxes checked) and delete all data and delete the bookmark. And no note that once I added back in all the other favicon links (.ico and .PNG versions) for other browser support, the Firefox bookmark favicon was the #2a2a2a bordered version. And no longer the SVG one. 
I could only get the Firefox bookmark favicon to stay the SVG one if there were no other links to any other favicon. Only the svg. 

I can have a look at the chromium one. I used a favicon generator website for the .ico and .PNG ones so maybe I'll try to not use those and replace them with actual versions that I create in gimp. I know I can output to .ico in gimp and .PNG is easy from the svg in inkscape.

TheHesster:
Update: I think I found out how to make sure chrome only uses the svg. I will experiment. If I find a fix, do I close this PR and submit another?

CGBassPlayer:
Nope! Just commit to your branch and GitHub handles the rest

TheHesster:
Oh wow ok. Thanks!

CGBassPlayer:
Just as a little explanation:

When you do a PR you are requesting your branch to be pulled into their code base. So until that pull actually happens, your branch can be updated and those changes after the PR is created are still part of the changes. That is why I can go in and request a change in your contribution and then still merge that same PR after that change has been made.

On GitHub, they treat it like a timeline so when you commit to your branch again, you will see your commit message(s) below this conversation. Hope this makes sense

TheHesster:
Yep makes sense!

TheHesster:
I think I've finally got it going. 
Here's screen shots of Firefox light/dark, bookmarks, tab and history.
DarkFirefox
LightFirefox

Here's Chrome light
Chrome

I can't get Chrome to switch themes for me on my laptop. Not sure why... Maybe someone can try with Dark Mode set on a Windows box?

gerbrent:
> is an old cached one.

RIGHT! forgot about that. Will clear everything and retest - good catch! Looks great in your above screenshots.

TheHesster:
I was able to reach the site on my windows machine and was able to get Chrome Dark mode screenshot.
ChromeDark

CGBassPlayer (approved):
So testing with Brave (Chromium), I can see the new icon works fine in both light and dark modes. 

image
(left tab is the local instance with the new favicon, right is the current production site)

I would like to see one more person review this before it gets merged.

Issue #232, opened 2022-08-06
Contact Page looks awful


reesericci:
The contact form page has no styling and is an iFrame of a 3rd-party service. It should be set up where the form is hosted on the JB site with all the JB styling and then submits the results with the `action` parameter to a 3rd-party API which processes the form and sends an email or whatever.

reesericci:
Why does the upload an image field exist if nobody is supposed to use it?

![image](https://user-images.githubusercontent.com/19589006/183251751-5b8e3f13-42d6-4b85-a494-38d1860c3271.png)

reesericci:
https://bulma.io/documentation/form/general/

gerbrent:
I love the frankness of this issue title ; )

reesericci:
Imo this is important for the 1.0 milestone to properly create a consistent contact form.

reesericci:
I'd be willing to work on it if I know what API I'm tying into.

ErraticKnight:
I'm guessing that would explain the weird behavior I saw yesterday there. If you drag the comment box's resizing corner you can pull far enough down to push the UI that lives below underneath the footer; if you pull enough you can get the hot zone for the resize corner pulled underneath the footer as well which makes the change unfixable.

elreydetoda:
Related to #139, the original plan that @kbondarev and I chatted about over there was to just have an API on the backend the form is submitted to. (so everything is self hosted)

We stopped pursuing that, because we had (at the time) though that the old contact form would be enough for the 1.0 release. Then we could do the re-write possibly for 2.0.

We're still only at 50% and there is still plenty to do, is there a critical reason why the 1.0 can't just re-use the old one? Since this is mainly an MVP (or Minimum Viable Product) for the 1.0 release, as long as people can still properly write in and submit feedback then IMO I think this could be put back to 2.0. ( trust me I know I'd love to add a bunch of things as well 😅)

The other thing to take into consideration is that if we change this, we might change the internal JB workflow. Which is probably a good thing to make it easier for them, than just consuming emails from a 3rd party. I think if we take our time on this and do it right, then we could make it really nice and easy for the JB team to get feedback. Making that all work for them by the launch deadline though, doesn't seem easily feasible (without it being someone dedicated to coordinate everything I mentioned above).

(Just my 2 cents)

elreydetoda:
> Why does the upload an image field exist if nobody is supposed to use it?
> 
[...]

PR #501, opened 2023-02-04
Schedules on the Calendar page

MarkSort:


MarkSort:
How it looks currently:

jupiter-calendar-schedule-1

content/show/coder-radio/_index.md
```diff
@@ -6,6 +6,8 @@ draft = false
 categories = ["Coder Radio"]
 show = "coder-radio"
 hosts = ["chris","michael", "wes"]
+live = "Mondays<br>9:00am Pacific<br>12:00pm Eastern"
```
MarkSort, line 9: I definitely don't plan to have HTML here.  Instead it will be some parse-able value we can use to display the time based on the chosen timezone.

themes/jb/layouts/calendar/single.html
```diff
@@ -5,11 +5,8 @@
   <div class="container">
       <div class="content">
         <h1>{{.Title}}</h1>
-        Subscribe To The Calendar: <a href="{{.Site.Params.calendar.rss}}">iCalendar</a>
-
-        <iframe frameborder="0" scrolling="no" data-src="//www.google.com/calendar/embed?showTitle=0&amp;showNav=0&amp;showPrint=0&amp;showTabs=0&amp;showCalendars=0&amp;mode=AGENDA&amp;wkst=1&amp;bgcolor=%23FFFFFF&amp;src={{.Site.Params.calendar.embedd}}&amp;color=%235229A3" width="100%" height="400px"></iframe>
-
-        <select name="ctz" id="ctz" style="width: 100%;">
+        <h2>Regular Live Show Times</h2>
```
MarkSort, line 8: I was thinking about how these shows have two schedules: a live show schedule and a recorded show release schedule.

Not sure how much this header helps differentiate but wanted to try something.

gerbrent:
Looks impressive on initial review - will give it some testing and look to some more talented folk to give comment on the implementation.

Thanks for this!

content/show/coder-radio/_index.md
```diff
@@ -14,6 +14,11 @@ header_image = "/images/shows/coder-radio.png"
 
 podverse_podcast_id = "ZXd_1Ojd9"
 
+[live]
+  type = "weekly"
+  day = 1
+  hour = 9
```
CGBassPlayer, line 20: The Coder Radio card seems to be ahead of the Google Calendar by about 3 hours.

image

themes/jb/layouts/calendar/single.html
```diff
@@ -320,24 +317,120 @@ <h1>{{.Title}}</h1>
...
+
+        Subscribe To The Calendar: <a href="{{.Site.Params.calendar.rss}}">iCalendar</a>
+
+        <iframe frameborder="0" scrolling="no" data-src="//www.google.com/calendar/embed?showTitle=0&amp;showNav=0&amp;showPrint=0&amp;showTabs=0&amp;showCalendars=0&amp;mode=AGENDA&amp;wkst=1&amp;bgcolor=%23FFFFFF&amp;src={{.Site.Params.calendar.embedd}}&amp;color=%235229A3" width="100%" height="400px"></iframe>
+
+
     </div>
   </div>
 
   <script>
```
CGBassPlayer, line 347: Since the JavaScript as increased significantly, I think it should be moved to its own JS file and imported to the page. 

Here is how the jb live stream handles its JavaScript

```html
{{ define "headcontent" }}
  {{ $jbTube := resources.Get "js/jb-live.js" }}
  {{ if hugo.IsProduction }}
    {{ $jbTube = $jbTube | minify | fingerprint }}
  {{ end }}
<script type="text/javascript" src="{{ $jbTube.Permalink }}" integrity="{{ $jbTube.Data.Integrity }}"></script>
```

themes/jb/layouts/partials/episode/schedule.html
```diff
@@ -0,0 +1,22 @@
...
+          data-live-schedule-day="{{ index .Params.live.day }}"
+          data-live-schedule-hour="{{ index .Params.live.hour }}"
+          {{ if (isset .Params.live "minute") }}
+            data-live-schedule-minute="{{ index .Params.live.minute }}"
+          {{ end }}
+      ></p>
+      </div>
+  </div>
+
+</div>
```
CGBassPlayer, line 22: Instead of multiple `a` tags for the image and the title, wrap the entire card in it so the entire card is a link to the show. This will match the functionality of other cards on the site

```suggestion
<a href="{{ .Permalink }}">
  <div class="card card-episode">
    <div class="card-image">
      <img src="{{.Params.header_image }}" width="432" height="242" alt="{{.Title}}">
    </div>
    <div class="card-content">
      <div class="content">
        <h3>{{.Title}}</h3>
        <p class="schedule"
           data-live-schedule-type="{{ index .Params.live.type }}"
           data-live-schedule-day="{{ index .Params.live.day }}"
           data-live-schedule-hour="{{ index .Params.live.hour }}"
           {{ if (isset .Params.live "minute") }}
             data-live-schedule-minute="{{ index .Params.live.minute }}"
          {{ end }}></p>
      </div>
    </div>
  </div>
</a>
```

CGBassPlayer (changes requested):
Overall a great (and I feel much needed) addition to the site! Just a few suggestions and tweaks to make this contribution that much better!

content/show/coder-radio/_index.md
MarkSort: I _think_ this is because today's time is not the regular time.  It should match if you look at the time for next week.  I can ask around too though.

But I guess it does bring up a non-technical question, is there a way we can make the reason these don't match more clear?  Not something we can do in the website, but could the calendar items add "(Special Time)" to the title, or something similar? 

(And I will take a swing at the other comments later, thanks!)

MarkSort: [Chris confirmed in matrix](https://matrix.to/#/!teHukPkJDFppjPjGQg:matrix.org/$Z0zqsiYcqdoFtIuh2iV-PklWcu5wPhm30XnfETMcVAE?via=matrix.org&via=jupiterbroadcasting.com&via=linuxdelta.com) that today's time is not the usual time.

CGBassPlayer: Looks like I should've looked farther ahead! I think we the calendar should have some kind of note if possible to specify that the time is different from normal. I haven't used Google Calendar so I am not sure if that is an option. We can ask Chris about that

CGBassPlayer: Another crazy thought... Could we use the Google Calendar API to get the time of the next event? Maybe outside the scope of this PR (unless you really want to tackle it). But that could be a good next step on this.

MarkSort: I wouldn't want to drag this out too much if the current version seems useful as a first iteration.

But separately, I wouldn't mind looking into using the Google Calendar API.  I'd guess the best way would be to call it during the Hugo build, instead of client side JS, so we don't have to worry about API keys being public?  If we could copy the next two months of showtimes into the static files, we could include "Next Showtime: ..." in each card.

For now, what if we put a header between the cards and the calendar that says something like "Upcoming Times - Includes Special Times"

CGBassPlayer: We are looking into how to handle the contact form with its API as well. I had proposed a separate backend just handle API requests like those. There was some discussion about it in #232

CGBassPlayer:
Initial scroll through the PR looks good. I will give a real review tomorrow

themes/jb/layouts/calendar/single.html
```diff
@@ -320,24 +326,30 @@ <h1>{{.Title}}</h1>
...
-        document.querySelector('iframe').src = calendarBase + "&ctz=" + tzname;
-      }
-    };
-  </script>
+        <br>
+        <br>
+
+        Subscribe To The Calendar: <a href="{{.Site.Params.calendar.rss}}">iCalendar</a>
+
+        <iframe frameborder="0" scrolling="no" data-src="//www.google.com/calendar/embed?showTitle=0&amp;showNav=0&amp;showPrint=0&amp;showTabs=0&amp;showCalendars=0&amp;mode=AGENDA&amp;wkst=1&amp;bgcolor=%23FFFFFF&amp;src={{.Site.Params.calendar.embedd}}&amp;color=%235229A3" width="100%" height="400px"></iframe>
```
CGBassPlayer, line 350: So personally I try to avoid `<br/>` for spacing content and untagged text in my html. I would use the classes provided by [Bulma](https://bulma.io/documentation/helpers/spacing-helpers/) to handle that with CSS. 

I am thinking this is fine for this PR. But a follow up (of either a calendar rework using or of a styling change) could be where this will happen.

CGBassPlayer (approved):
I think this PR is good as is for now. But see my comment about a future PR.

In local testing:
- dev container built and ran
- basic click through testing looked good
- e2e tests passed

themes/jb/layouts/calendar/single.html
MarkSort: I added a div and some Bulma classes and removed the `<br>`s.

content/show/coder-radio/_index.md
reclaimingmytime: > so we don't have to worry about API keys being public?

We can look into settings to protect and scope the Google Calendar API key properly. Such as only allowing access to the JB Calendar from jupiterbroadcasting.com, and nothing else from Google Calendar. I haven't worked with the Google Calendar API, but I've seen other APIs allow these kinds of settings, so it should work with Google Calendar as well. Quickly glancing at the API, it seems to be possible with https://cloud.google.com/docs/authentication/api-keys#api_key_restrictions.

I see https://irc.twit.tv/ is using a solution based on the Google Calendar API, we might be able to build something similar. You can search for "calendar" in the Network tab of your browser and see what Google API they are using. I'm hoping they've done their research on not expoing their API key.

If we have an API key, this might just be a separate AJAX request to a JSON API, similar to what TWiT is using.

> I wouldn't want to drag this out too much if the current version seems useful as a first iteration.

I agree, but I think if we change the calendar, we should make sure it's correct and not too confusing. I think the design of the cards looks great, and it's definitely a great addition. Just hard-coding the events might make it more confusing. A better solution could be using the Google Calendar API, in my opinion.

MarkSort: I did get excited to have something merged here, but makes sense to back up and reassess.  😅

Are we still talking about adding "Next Showtime" to the cards?  And potentially using the Google Calendar API to determine that?  

We'd still need to indicate when the "Next Showtime" is not a regular show time.  We could maybe do that by calculating when the next regular time _should_ be, and if they don't match, we know it's special.

reclaimingmytime: > We could maybe do that by calculating when the next regular time should be, and if they don't match, we know it's special.

That could work. Maybe looking at the last 3-4 shows and determining a pattern could be an idea.

We should also take permanent changes in the live calendar into account, like how LUP was moved from Tuesday to Sunday.

I also noticed that some event in the calendar have a different name to make it clear it was a different live time.

image

Maybe this will make it more difficult to automatically detect which show is live? We could detect any show with "Coder Radio" in the title, but what about false positives like "No Coder Radio this week"? Have we had this before? 😁

Or maybe it's good enough to have a system that works most of the time? Google has changed the API before, and there will be changes in the future, so there might always be unforeseen breakages.

MarkSort: > "No Coder Radio this week"

Nevermind I give up.

Just kidding, but really my goal here was to have somewhere on the website to find general live schedules for each show.  We immediately hit a corner case where it was confusing, so it became more difficult to proceed.  If I have time I'll check out the Google Calendar API.

> Maybe looking at the last 3-4 shows and determining a pattern could be an idea.

I'd hope we can be more precise, either by looking at the recurrence of the Google Calendar events, or having something like I have in the `live` property of shows in this PR.

PR #502, opened 2023-02-05
Temp Fix for Contact Page

### LINKED ISSUE #485: Contact - submit button hidden on missing field
From a LUP listener, and will likely need more investigation:

> Btw the submit button is not reachable, when missing a mandatory field and the notification is shown, for me on mobile with my pixel7pro. Don't know if it's resolution related but I had to put in everything again to send it.

CGBassPlayer:
Fixes #485 

On Safari/Webkit browsers, the iframe was too small and cut off the bottom button when all validation errors are present. This increases the size of the iframe by 5 pixels as a band-aid until we remake the contact form.

e2e testing for that browser is part of #475

### DIFF
--- a/themes/jb/layouts/contact/single.html
+++ b/themes/jb/layouts/contact/single.html
@@ -6,7 +6,9 @@
   <div class="container">
     <div class="content">
       <h1>{{.Title}}</h1>
-      <iframe id="contact-frame" height="830" allowTransparency="true" frameborder="0" scrolling="no" style="width:100%;border:none"  src="https://jblive.wufoo.com/embed/w7x2r7/"><a target="_blank" href="https://jblive.wufoo.com/forms/w7x2r7/">Fill out the form here</a></iframe>
+      <iframe id="contact-frame" height="835" allowTransparency="true" frameborder="0" scrolling="no" style="width:100%;border:none"  src="https://jblive.wufoo.com/embed/w7x2r7/">
+        <a target="_blank" href="https://jblive.wufoo.com/forms/w7x2r7/">Fill out the form here</a>
+      </iframe>
     </div>
   </div>
 {{end}}


gerbrent (approved):
tested, looks good as expected.

gerbrent:
proof from Firefox' responsive mode (though I can't reproduce the issue on current production JB site w this method......):

image

gerbrent:
> for webkit browsers

The [original issue](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/485#issue-1447018809) was reported by a user of a "pixel7pro"

felt important to point this out.. and curious if the fix might influence that context aswell - likely!

CGBassPlayer:
It does and it doesn't. The test expansion handles multiple browsers, but I haven't worked in emulation yet. But the WebKit browser failed the test without this change. 

So while it will it fixes all the browser at a large scale. I can't say for certain it fixes the mobile simply because I don't have a way to test yet.

gerbrent:
.... you do now!

https://jb-deploy-prv-iac8.onrender.com/

CGBassPlayer:
So on my iPhoneX it seems like it wasn't quite enough... I am not too worried about it since I want to get a native contact form working on the site soon.
image

gerbrent:
Want to give it another 5-10px while we're at it?

CGBassPlayer:
I was thinking at least 20px. Since this PR is merged, I am pretty sure I would need to open a new one

gerbrent:
other way is a revert (I can do that) and change, then remerge?

That would have a slightly cleaner history.. I believe.

CGBassPlayer:
That actually could work since I don't think there was anything merged after this one. We can certainly give it a shot

gerbrent:
note that... revert doesn't necessarily help w clean history - it's subjective at this point ; )

gerbrent:
@CGBassPlayer I'll let you do the new PR w above tiny changes  - probably overkill to revert at this point but... today we learn.

PR #504, opened 2023-02-06
Increase Contact iframe

CGBassPlayer:
Increase Contact iframe to 1000px to hopefully be able to all validation errors on mobile screen without submit button disappearing.


gerbrent (approved):
lgtm

PR #505, opened 2023-02-06
Syncing main to develop

CGBassPlayer:
Add latest episodes and people to the develop branch


gerbrent (approved):
looks good and simple!

elreydetoda:
So, in the future (if someone with write permissions) we don't have to create a PR for this. If we do a sync from dev to main first (i.e. #509 ), then you can run `git pull --ff-only origin main` on the develop branch and just push from your local to the develop branch. Then it'll bring in ALL new changes and be identical to main.

PR #507, opened 2023-02-13
Replace space with - in tags url

### LINKED ISSUE #497: 404 when accessing tags with a space
### URL

https://www.jupiterbroadcasting.com/show/linux-unplugged/494/

Phone, Desktop

Chrome

### Summary

When I access tags with multiple words, such as "jupiter broadcasting" or "remote desktop", I get redirected to a 404 page.
image
image

Tags with one word redirect to the proper tags overview.

Maybe the issue is related to how a space is converted?

Balastrong:
This PR fixes #497.

Hello everyone! I'm kind of new here but I'm looking for new challenges so here we are :)

I noticed [`urlize`](https://gohugo.io/functions/urlize/) has been replaced with [`urlquery`](https://gohugo.io/functions/urlquery/) in #456 (Issue #319).

However, with this change the tag `c#` works, but tags with space in between don't (they did with `urlize` though).

I did some attemps and searches but couldn't find anything 100% working so the idea I had to keep the best of both worlds is to replace the empty space ` ` with `-` first (so that tags with the space are ok) and then keep urlquery for weird chars like `#` and `+`.

Tested with:
- `c# -> c%23` ok
- `c++ -> c%2B%2B` ok
- `csam scanning -> csam-scanning` ok
- `🦒 -> %F0%9F%A6%92` not ok :(

The giraffe emoji is escaped just fine, the final url is `/tags/🦒/` which I would say is correct but I can't even manually find the correct tag page. Why would you use an emoji as tag anyway? 😂

### DIFF
--- a/themes/jb/layouts/partials/episode/tags.html
+++ b/themes/jb/layouts/partials/episode/tags.html
@@ -1,6 +1,6 @@
 {{ range $tag := . }}
   <span class="tag">
     <!-- urlquery found here: https://discourse.gohugo.io/t/url-encoding-percent-encoding-with-hugo/16546/13 -->
-    <a href="{{ `/tags/` }}{{ urlquery $tag }}/">{{$tag}}</a>
+    <a href="{{ `/tags/` }}{{ replace $tag " " "-" | urlquery }}/">{{$tag}}</a>
   </span>
 {{ end }}
--- a/test/e2e/test_edge-cases.py
+++ b/test/e2e/test_edge-cases.py
@@ -1,0 +1,9 @@
+from playwright.sync_api import Page, expect
+
+"""
+Checking for broken tag links with spaces in them
+"""
+def test_tag_broken_link_spaces(page: Page):
+    page.goto("/show/linux-unplugged/489/")
+    page.locator(".tag > a", has_text="linux unplugged").click()
+    expect(page.locator(".title", has_text="Tag: linux unplugged")).to_be_visible()


reclaimingmytime:
> Why would you use an emoji as tag anyway? 😂

I assume they used a 🦒 because of the inside joke "giraffe-ene OS" instead of "GrapheneOS" 😄 

I noticed that in the markdown, the emoji `🦒` gets rendered as `\ud83e\udd92`.
https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/581c099b841a0d9f8a2af80ceb491c6143146beb/content/show/linux-unplugged/0489.md?plain=1#L75

I don't know enough about Hugo to provide a solution, but maybe the function _emojify_ could be helpful (turn an emoji into text, then turn it back to an emoji when it's displayed).
https://gohugo.io/functions/emojify/

themes/jb/layouts/partials/episode/tags.html
```diff
@@ -1,6 +1,6 @@
 {{ range $tag := . }}
   <span class="tag">
     <!-- urlquery found here: https://discourse.gohugo.io/t/url-encoding-percent-encoding-with-hugo/16546/13 -->
-    <a href="{{ `/tags/` }}{{ urlquery $tag }}/">{{$tag}}</a>
+    <a href="{{ `/tags/` }}{{ replace $tag " " "-" | urlquery }}/">{{$tag}}</a>
```
reclaimingmytime, line 4: Is there a reason why you used `urlquery` instead of `urlize`? This function seems to replace hyphens with spaces internally, so the syntax is cleaner to read.
https://gohugo.io/functions/urlize/

Balastrong: `urlquery` won in favor of `urlize` in #319

elreydetoda (changes requested):
Hey @Balastrong,

First off, thank you for submitting a PR to fix this issue! We really appreciate it :partying_face: 

Secondly, everything looks good to me (LGTM) and I will approve the PR but I wanted us to have a test in place so we can ensure this doesn't happen again. I created a test and submitted a PR against you PR's branch (https://github.com/Balastrong/jupiterbroadcasting.com/pull/1).

If you wouldn't mind accepting that I'll happily merge the changes in afterwards :grin: 

Thank you again for your contribution and I look forward to getting this fixed soon! :partying_face: :tada:

Balastrong:
Hey @elreydetoda, thank you for the review and for adding the test!

I merged your PR now :)

elreydetoda (approved):
Awesome, thanks @Balastrong !

LGTM

PR #544, opened 2023-06-26
Set metadata properly for hosts and guests

### LINKED ISSUE #520: Host/Guest Pages use usernames in Browser Tab
### URL

https://www.jupiterbroadcasting.com/hosts/wes/  (But all host and guest pages)

Desktop

Chrome

### Other Browser

_No response_

### Summary

Host and Guest pages use the `username` in the page title which is the text for the tab in a desktop/tablet browser. As far as I can tell this is just a placeholder username to get everyone over to the new site. This should be updated to use the `title` field as that seems the be there actual name.

image
image

FryDay:
Because people are setup as taxonomies, a little more work needs to be done in order to properly set the page metadata.

Fixes #520

### DIFF
--- a/themes/jb/layouts/partials/meta.html
+++ b/themes/jb/layouts/partials/meta.html
@@ -7,6 +7,19 @@
     -}}
         {{/* for all other shows */ -}}
         {{- $titleMeta = (printf `%s | %s %v | %s` .Title .Params.show_name .Params.episode .Site.Title) -}}
+    {{ else if or
+        (eq .Type `hosts`)
+        (eq .Type `guests`)
+    -}}
+        {{/* for people */ -}}
+        {{
+          $person := index (
+            where (
+              where site.RegularPages "Section" "people"
+            ) ".Params.username" .Title
+          ) 0
+        }}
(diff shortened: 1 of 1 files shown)


CGBassPlayer (changes requested):
First off: Thank you for contributing to the project! :tada: 

When testing this locally, I can see that the title listing the main [hosts](https://www.jupiterbroadcasting.com/hosts/) and [guests](https://www.jupiterbroadcasting.com/guests/) are 
`%!s(<nil>) | Jupiter Broadcasting`

image

image

My best guess would be that you have to update your if condition to only check actual host and guest pages vs the main lists. Not sure how exactly that would look since I have not been in the code recently.

CGBassPlayer (approved):
Much better!

image

image

Screenshots showing names in the tab title
image

image

Thank you again for this! @FryDay

FryDay:
> Thank you again for this! @FryDay

No problem! I've been a religious listener to every JB podcast since 2011. Figure it's about time I give back. I'm sure this won't be my last contribution 😄

CGBassPlayer:
I know the E2E test failed. Something weird is going on with the iframe loading times in the testing and that is causing the test to fail. I will merge this and hopefully look into the soon

PR #546, opened 2023-07-07
Update matrix page

kbondarev:
I created a new matrix account and had to rejoin the JB spaces. When I used the matrix page for reference I got a little confused and originally thought that the `community` space is the main one (probably because it was the first link in the list and the comment that was near the link) and I missed the link to the main JB space :confused: 

I suspect that I might not be the only one who will find it confusing :sweat_smile: , and therefore I made these slight modifications which I believe make things a little clearer visually

### DIFF
--- a/content/community/matrix.md
+++ b/content/community/matrix.md
@@ -15,16 +15,15 @@ aliases = [
 
 ## Spaces and Rooms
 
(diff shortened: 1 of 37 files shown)


kbondarev:
Oops I made a mess here. I'm gonna fix this and make a new PR

PR #547, opened 2023-07-07
Update matrix page to make the spaces structure clearer

kbondarev:
I created a new matrix account and had to rejoin the JB spaces. When I used the matrix page for reference I got a little confused and originally thought that the community space is the main one (probably because it was the first link in the list and the comment that was near the link) and I missed the link to the main JB space :confused: 

I suspect that I might not be the only one who will find it confusing :sweat_smile:  , and therefore I made these slight modifications which I believe make things a little clearer visually

### DIFF
--- a/content/community/matrix.md
+++ b/content/community/matrix.md
@@ -15,16 +15,15 @@ aliases = [
 
 ## Spaces and Rooms
 
-Jupiter Broadcasting's Rooms are organized into groups called "spaces".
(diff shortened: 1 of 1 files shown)


CGBassPlayer (approved):
LGTM. I think this much better explains the matrix room structure

CGBassPlayer:
Tests are failing on the live page indicator which this change is not affected by. I will be looking into why it is failing soon

PR #553, opened 2023-08-24
Add missing direct download links to back catalog

ChanceM:
Closes JupiterBroadcasting/show-scraper#29

Once JupiterBroadcasting/network#20 is resolved we can add the final missing link.


ChanceM:
I slightly misunderstood the problem based on the other issue linked. This resolves the missing info for the episodes that where caught in between the go live of the new site. JupiterBroadcasting/show-scraper#29 will probably need to remain open, as I now understand it is searching for a replacement source of information for post go live shows rather than those caught in the middle.

CGBassPlayer:
There is a lot of content to review in this one. I will hopefully look through it all tomorrow night

content/show/coder-radio/0476.md
```diff
@@ -55,11 +55,11 @@
   "podcast_chapters": null,
   "podcast_alt_file": null,
   "podcast_ogg_file": null,
-  "video_file": null,
+  "video_file": "https://201406.jb-dl.cdn.scaleengine.net/coderradio/2022/cr-0476.mp4",
```
reclaimingmytime, line 58: This link is not accessible for me, so something must have changed in the last two weeks? Or maybe it's just temporary?

content/show/coder-radio/0476.md
```diff
@@ -55,11 +55,11 @@
...
   "video_hd_file": null,
   "video_mobile_file": null,
-  "youtube_link": null,
-  "jb_url": null,
+  "youtube_link": "https://youtu.be/dscw9-81RfA",
+  "jb_url": "/149367/tapping-the-breaks-coder-radio-476/index.html",
```
reclaimingmytime, line 62: Why do we need this? The old WordPress site is no longer up. Or should this link to https://original.jupiterbroadcasting.net/149367/tapping-the-breaks-coder-radio-476/index.html?

ChanceM: It's not used but does provide context to where this information came from for historical purposes.

content/show/coder-radio/0476.md
CGBassPlayer: @ChanceM would you be able to update these Scale Engine links the HTTP? I think that makes the most sense for time being.

content/show/coder-radio/0476.md
CGBassPlayer: I think leaving them in for historical purposes will be find. Leaving the base URL off makes sense as well because in the off chance that the guys decide they want to host the old site somewhere else, we are not tied to that domain.

CGBassPlayer (approved):
LGTM

Issue #534, opened 2023-03-26
404 when accessing tags with an apostrophe (’)


CGBassPlayer:
### URL

https://www.jupiterbroadcasting.com//tags/conan-o%E2%80%99brien-podcast/ ([OH 5](https://www.jupiterbroadcasting.com/show/office-hours/5/))
https://www.jupiterbroadcasting.com/tags/department-of-commerce%E2%80%99s-national-telecommunications-and-information-administration/ ([CODER 504](https://www.jupiterbroadcasting.com/show/coder-radio/504/))

### On what type of device(s) did you see this bug?

Phone, Tablet, Desktop

### On which browser(s) are you seeing the problem?

Firefox, Chrome, Safari, Microsoft Edge

### Summary

Similar issue to #497
While working on e2e testing. I noticed that the `conan o’brien podcast` tag from Office Hours 5 results in a 404. I am assuming this is from the apostrophe causing some kind of issue with generating the files possibly?

reclaimingmytime:
This could also be related to issue #516.

CGBassPlayer:
I think it is close. It seems like they both suffer from differing between the URL for the tag `\ud83e\udd92` vs :giraffe: and `conan-o%E2%80%99brien-podcast` vs `conan o’brien podcast`

CGBassPlayer:
Update. I built the site locally and the tag page is `conan-obrien-podcast`... So that means that https://www.jupiterbroadcasting.com/tags/conan-obrien-podcast/ is the correct URL and hugo is removing apostrophes from tags when it builds the HTML files.

PR #554, opened 2023-08-28
Add listing of all tags

### LINKED ISSUE #500: "Tags" url/page present no content. Regression?
> [...]the "Tags" url seems.... broken aswell. In my recollection, it used to present all tags ever present in alphabetical order, however now simply presents a "Tags" title w no content.

https://www.jupiterbroadcasting.com/tags/

_Originally posted by @gerbrent in https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/issues/497#issuecomment-1416876932_

ChanceM:
Closes #500. Adds list of tags to /tags url

### DIFF
--- a/themes/jb/layouts/_default/terms.html
+++ b/themes/jb/layouts/_default/terms.html
@@ -14,7 +14,7 @@
       {{ range sort .Site.Taxonomies.tags }}
       <span class="tag">
         <!-- urlquery found here: https://discourse.gohugo.io/t/url-encoding-percent-encoding-with-hugo/16546/13 -->
-        <a href="{{ .Page.Permalink }}">{{ .Page.Title }} </a>
+        <a href="{{ .Page.Permalink }}">{{ .Page.Title }} ({{ .Count }})</a>
         </span>
 
       {{ end }}
--- a/themes/jb/layouts/_default/terms.html
+++ b/themes/jb/layouts/_default/terms.html
@@ -1,0 +1,28 @@
+{{ define "main" }}
+
+<div class="px-4 container">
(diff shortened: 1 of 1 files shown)


ChanceM:
Screenshot_11

CGBassPlayer:
Is there a way to get a count of the episodes that use that tag? I'm thinking about ways to clean up the page and gaveling a count with the tag would be a nice addition. Even if you leave that code commented out for now.

Regardless, thank you for working on this. I will hopefully review it some time today

ChanceM:
> Is there a way to get a count of the episodes that use that tag? I'm thinking about ways to clean up the page and gaveling a count with the tag would be a nice addition. Even if you leave that code commented out for now.
> 
> Regardless, thank you for working on this. I will hopefully review it some time today

Actually I had that in there and must have taken it out while playing around with it. I'll add it back.

CGBassPlayer:
I have not looked at the code yet (only on mobile currently). If it is there, awesome, if not, just add it back in so myself or someone who actually knows website design can have that data.

ChanceM:
image

CGBassPlayer (approved):
This is definitely a great start! But I think that we should open up an issue to improve the design of the page a little bit... This wall of tags can feel like a bit much when you are looking through. 

I also caught that `/tags/#yesmom` does not load. I assume it is trying to anchor itself to somewhere on the tags page that does not exist. I will reopen #534 and add this new information I have discovered.

ChanceM:
Yeah they don't handle the # as it is technically allowed in the URL. If we manually replace with "%23" they will work. I can fix that.

Issue #377, opened 2022-09-02
hosts page - incorrect layout after photo change


gerbrent:
Alex changed his profile photo in #373 
This seems to have caused layout issues 

![image](https://user-images.githubusercontent.com/7018928/188037073-02b3a664-89a6-45aa-9ec1-2d5c9a4ab3a1.png)

@ironicbadger can you tell us where you got the photo? I presume you copied it from some other online profile where the image has been shrunk.

My theory:
I presume given it's data size - 29 KB - and it's pixel dimensions - 300x263 - that a too-small image size applied to a profile causes layout problems instead of being scaled/interpolated (despite pixelation).

@ironicbadger can you procure a higher-resolution copy relatively easily? Then again, this may not be the solution, but is a bandaid...

elreydetoda:
@ChanceM (guessing that's them, since same name in matrix) posted this:

>  Looking at issue #377, which led down a deeper rabbit hole. Do you know why hosts/guests are not rendered with the people layout? Right now they are rendered with the default list layout, not sure if they was by design.

my brief response (more later):

> tl;dr: IIRC since there is no content in the content/hosts/ (& guests) then there is no way to apply a list page to it besides through the default list page (at least from what I could tell)
>
>So, that's why I [check for hosts/guests](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/24a2507059aea540778dbef0c740ffac4be0aba9/themes/jb/layouts/_default/list.html#L14)
>
>Plus the people's page lists both out, so it's got an [extra range](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/24a2507059aea540778dbef0c740ffac4be0aba9/themes/jb/layouts/people/list.html#L14) for the groups of people.

ChanceM:
> @ChanceM (guessing that's them, since same name in matrix) posted this:
> 
> >  Looking at issue #377, which led down a deeper rabbit hole. Do you know why hosts/guests are not rendered with the people layout? Right now they are rendered with the default list layout, not sure if they was by design.
> 
> my brief response (more later):
> 
> > tl;dr: IIRC since there is no content in the content/hosts/ (& guests) then there is no way to apply a list page to it besides through the default list page (at least from what I could tell)
> >
[...]

PR #556, opened 2023-09-02
Expand clickable area, add gradient to define it

### LINKED ISSUE #186: home carousel - scroll arrow UX improvement
reference/history: #174 

I have a recommendation for UX improvement on the home page carousel:

#### Current behaviour:
* click the `>` arrow to scroll right. Since we don't know how many shows there are, we keep clicking... two clicks reaches the end of the carousel, third click registers as a click on Self Hosted Show, since the arrow vanishes before you really realize it.

#### Better behaviour:
Navigating the carousel should not inadvertently take us away from the home page.

The carousel arrows could perhaps, as options:
* be independent of the shows i.e. not overlay them
* the arrow could persist, even when the end of the carousel is reached, thereby not allowing additional clicks to follow to a show (unless the mouse is moved off the arrow, of course...)

thoughts @reesericci ?

Overall, I love the idea of the carousel - really adds to the homepage and make for quick, fun navigation. Totally 🎉  on that idea!

### LINKED ISSUE #530: Carousel to open episode on Fire Tab 7
### URL

_No response_

Tablet

Chrome

### Other Browser

Amazon Silk

### Summary

I see 1.5 of carousel icons in portrait view and click on the 0.5 icon which advances that icon to left justification instead of opening it as i expected.
when i click the 0.5 icon, i expect it to open. i must be clicking the advance arrow, so the icon advances instead of opens. i wanted it to open. when it advances to left justified, i am able to open it, but this is an extra step. i'd rather it opened when i clicked it the first time than have it advance where i have to click it again to open it.
this has something to do with the resolution of the Fire Tab 7
as discussed on matrix earlier today

ChanceM:
Fixes #186  and Fixes #530 by expanding the handles to the edges of the carousel and adds gradients to help establish the areas.

### DIFF
--- a/themes/jb/assets/css/layout/show-slider.scss
+++ b/themes/jb/assets/css/layout/show-slider.scss
@@ -1,7 +1,7 @@
 
 .show-img {
-//  height: 168px !important;
-//  width: 300px !important
+//  height: 170px !important;
+//  width: 303px !important
 }
 
 .show-card {
--- a/themes/jb/assets/css/layout/show-slider.scss
+++ b/themes/jb/assets/css/layout/show-slider.scss
@@ -1,7 +1,7 @@
 
 .show-img {
- height: 168px !important;
- width: 300px !important
+//  height: 168px !important;
+//  width: 300px !important
 }
 
 .show-card {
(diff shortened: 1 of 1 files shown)


ChanceM:
PR:
image

Original: 
image

CGBassPlayer:
Just for reference:

Actual appearance on a "light" background
image

Actual apperance on a "dark" background
image

themes/jb/assets/css/layout/show-slider.scss
```diff
@@ -1,7 +1,7 @@
 
 .show-img {
- height: 168px !important;
- width: 300px !important
+//  height: 168px !important;
+//  width: 300px !important
 }
```
CGBassPlayer, line 5: Any reason for removing this? I know the show's images are about that size anyway, but I wonder if we leave the sizing there just in case a new show gets added at some point in the future and the size does not match the rest.

What do you think?

ChanceM: Because it actually creates dead pixel space and will cause the alignment to look off. The images will fill their containers. Unless the aspect ratio changes (which would be a bigger issues) I don't see much risk.

CGBassPlayer: I just think back to the issue we had with profile pictures (#377). Just want to be prepared for as much as possible. Looking at the images, they seem to resize themselves a few pixels larger:
image

Maybe just update the comments to match what the images are when they fill the containers? That way if we ever need to bring them back, it'll be ready to go.

CGBassPlayer (approved):

PR #557, opened 2023-09-03
Mobile: move sponsors, clean padding, condense host/guest list

### LINKED ISSUE #430: Episode Page - move Sponsors above Episode Links in mobile view
### URL

everywhere! ; ) All episode pages in mobile view

Phone

Firefox

### Other Browser

Firefox' Responsive Design Mode

### Summary

On the Episode pages, it would be preferably to have the Sponsors section above the Episode Links section.

related to #400 

image

### Contributing

- [X] I have read through the [Contributing Guidelines](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/main/CONTRIBUTING.md)

ChanceM:
Closes #430, adjusts padding and condenses host/guest list layout on mobile.

PR:
localhost_1313_show_linux-unplugged_513_(iPhone XR)

Original:
www jupiterbroadcasting com_show_linux-unplugged_513_(iPhone XR)

### DIFF
--- a/themes/jb/assets/css/main.sass
+++ b/themes/jb/assets/css/main.sass
@@ -7,3 +7,10 @@
 @import "helpers/_all"
 #wrapper
   padding-top: 4rem
+
+.host-list .columns
+  @include until(600px)
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    .column:first-child:last-child
+      grid-column: 1 / span 2;
--- a/themes/jb/assets/css/main.sass
+++ b/themes/jb/assets/css/main.sass
@@ -10,7 +10,7 @@
 
 .host-list .columns
   @include until(600px)
-    display: grid;
-    grid-template-columns: repeat(2, 1fr);
+    display: grid
+    grid-template-columns: repeat(2, 1fr)
     .column:first-child:last-child
-      grid-column: 1 / span 2;
+      grid-column: 1 / span 2
--- a/themes/jb/layouts/episode/single.html
+++ b/themes/jb/layouts/episode/single.html
@@ -39,8 +39,8 @@
       <div class="tile is-ancestor mt-5">
         <div class="tile is-parent is-vertical is-8">
           <!-- Hosts Section -->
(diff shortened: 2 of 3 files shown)


themes/jb/layouts/episode/single.html
```diff
@@ -71,20 +71,37 @@ <h2 class="has-text-centered-mobile">Guests</h2>
...
+                    {{ $sponsor := index (where (where site.Pages "Section" "sponsors") "Params.shortname" .) 0 }}
+                    {{ if $sponsor }}
+                      {{ partial "sponsor/small.html" $sponsor }}
+                      <br />
+                    {{ end }}
+                  {{ end }}
+                </section>
+              {{ end }}
+            </div>
+          </div>
```
CGBassPlayer, line 90: Interesting approach! I was trying to figure out a way to change the order of the `div`s based on view port size. But I could not get that to work since they were in their own columns. I think this solution is honestly the simplest for what we are trying to accomplish. Maybe just throw a comment at each sponsor section to say why we have it defined twice in the HTML.

ChanceM: Comment added. Yeah without using something like css grid or JS its not really possible.

themes/jb/assets/css/main.sass
```diff
@@ -7,3 +7,10 @@
 @import "helpers/_all"
 #wrapper
   padding-top: 4rem
+
+.host-list .columns
+  @include until(600px)
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    .column:first-child:last-child
+      grid-column: 1 / span 2;
```
CGBassPlayer, line 16: SASS does not use semi colons on the ends of the lines. While I don't think it breaks anything, they should be removed to match the rest of the stylesheets
```suggestion
    display: grid
    grid-template-columns: repeat(2, 1fr)
    .column:first-child:last-child
      grid-column: 1 / span 2
```

CGBassPlayer (changes requested):
Just a minor syntax fix and this will be ready to go!

CGBassPlayer (approved):

PR #562, opened 2023-09-06
Replace # with urlEncoded %23

### LINKED ISSUE #559: Tags with `#` Do Not Load
### URL

https://www.jupiterbroadcasting.com/tags/#yesmom

Desktop

Chrome

### Other Browser

_No response_

### Summary

When reviewing #554, I noticed that `/tags/#yesmom` does not actually load that tags page. Due to the URL contain a `#`, the browser attempts to go to the anchor on the tags page that does not exists, making it look like nothing is happening when the link is clicked.

ChanceM:
Resolves #559


CGBassPlayer (approved):
Simple and elegant solution. Working locally:
image

PR #564, opened 2023-09-10
Push footer to the bottom of page

sgvictorino:


CGBassPlayer (approved):
This is actually really helpful when there is not a full page of content displayed

Issue #558, opened 2023-09-04
Update Design of Tags Page


CGBassPlayer:
Since #554 takes all the tags and dumps them on a single page, I think this page could use a little bit of love on the design side. Would love to see what idea people have for this!

CGBassPlayer:
From @reclaimingmytime 
> You can't really read through all tags at once. [The site](https://www.jupiterbroadcasting.com/tags/) has over 12,212 tags, 9,280 of them are just for one show. Most users are just going to see a relatively small part of the tags anyway.
>
>I think we should leave this choice up to the user. For reading related tags such as "zfs", "zfs 2.1...", etc., alphabetical notation makes sense. But for discovering new shows and getting a sense of what they're about, sorting by count is preferable.
>
>So, maybe create two tags pages, one for each sorting method?

I do agree that we will probably need to have some way to gives the users a way to look through the way we want. But I am also thinking a wall of tags probably isn't the best either. I wam thinking we should have some kind of divider (By first character for alphabetical and maybe the number of times its used for the popularity sort). 

I can't remember if it was specifically said, but both options really should to be available on the same page if possible. I just don't think it makes sense to have an alphabetic version and a popularity version unless we have some sort of technical limitation.

ChanceM:
Here is an alternative that is tags by show. Still alpha sort and counted per show. It takes about a minute to render though.
![localhost_1313_tags_](https://github.com/user-attachments/assets/2f943900-5bf9-4b80-b74a-040e65d843cc)

ChanceM:
Here's with a top 10: 
![image](https://github.com/user-attachments/assets/84041c63-7697-441c-97a2-ebec72899246)

PR #565, opened 2023-09-10
Sort the all tags list by count

sgvictorino:


CGBassPlayer:
Curious how others feel about about ordering the tags from most used to least used. 

I personally think it is fine because when I would normally Ctl + F the page and search for the tag I am looking for. My concern is that mobile users or anyone who accesses the page with a browser that does not have an easy search option would struggle if they are looking for a specific tag.

reclaimingmytime:
You can't really read through all tags at once. [The site](https://www.jupiterbroadcasting.com/tags/) has over 12,212 tags, 9,280 of them are just for one show.

I think we should leave this choice up to the user. For reading related tags such as "zfs", "zfs 2.1...", etc., alphabetical notation makes sense. But for discovering new shows and getting a sense of what they're about, sorting by count is preferable.

So, maybe create two tags pages, one for each sorting method?

CGBassPlayer:
We actually have #558 open for reworking the tags. I am going to close this PR as I don't think it makes the most sense to sort by most commonly used by default. But we can work on a new solution over there.

PR #567, opened 2023-09-11
Update Twitter to X branding

ChanceM:
image

### DIFF
--- a/content/community/twitter.md
+++ b/content/community/twitter.md
@@ -1,10 +1,10 @@
 +++
-title = "Twitter"
+title = "X"
 # https://en.wikipedia.org/wiki/Twitter
-description = "Twitter is a microblogging and social networking service on which users post and interact with messages known as \"tweets\""
+description = "X, is a microblogging and social networking service on which users post and interact with messages known as \"tweets\""
 date = "2022-07-23T03:51:55-04:00"
 draft = false
-logo = "https://upload.wikimedia.org/wikipedia/commons/4/4f/Twitter-logo.svg"
+logo = "https://upload.wikimedia.org/wikipedia/commons/c/ce/X_logo_2023.svg"
 comms_type = "network"
-direct_link = "https://twitter.com/jupitersignal"
+direct_link = "https://x.com/jupitersignal"
 +++
--- a/themes/jb/layouts/partials/episode/subscribe-dropdown.html
+++ b/themes/jb/layouts/partials/episode/subscribe-dropdown.html
@@ -1,0 +1,24 @@
+<div class="dropdown is-info py-2">
+    <div class="dropdown-trigger">
+      <button class="button is-info" aria-haspopup="true" aria-controls="dropdown-menu">
+        <span>Subscribe to {{.show}}</span>
+      </button>
+    </div>
+    <div class="dropdown-menu" id="dropdown-menu" role="menu">
+      <div class="dropdown-content">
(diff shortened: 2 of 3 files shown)


reclaimingmytime:
Maybe a white X is more readable on a dark background?

themes/jb/layouts/partials/episode/subscribe-dropdown.html
```diff
@@ -0,0 +1,24 @@
...
+        {{- range $key, $url := .direct_links -}}
+        <a href="$url" class="dropdown-item" style="text-transform: capitalize;">
+          {{- replace $key "_" " " -}}
+        </a>
+        {{- end -}}
+      </div>
+    </div>
+  </div>
+  <script>
+    var dropdown = document.querySelector('.dropdown button');
```
CGBassPlayer, line 19:
```suggestion
    const dropdown = document.querySelector('.dropdown button');
```

themes/jb/layouts/partials/episode/subscribe-dropdown.html
CGBassPlayer, line 1: I may be missing something but I do not see where this new partial is being used

CGBassPlayer (changes requested):
Overall looks great. But I will say that this PR should include updating the Twitter logo from the show pages as well. That should be as easy as adding another else if to the [links.html](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/develop/themes/jb/layouts/partials/show/links.html) file.
image

ChanceM:
> Overall looks great. But I will say that this PR should include updating the Twitter logo from the show pages as well. That should be as easy as adding another else if to the [links.html](https://github.com/JupiterBroadcasting/jupiterbroadcasting.com/blob/develop/themes/jb/layouts/partials/show/links.html) file.
> image
> 

Unfortunately I think that needs to be its own PR. Those icons are from font awesome of which the theme uses Fork Awesome which does not yet include an icon for X. An update to that library would also warrant testing of all icons used on the site.

This change fixes what is now a broken image on the site.

themes/jb/layouts/partials/episode/subscribe-dropdown.html
ChanceM: You are correct this was part of another change I was working on to add a drop-down with links to all the services to sub scribe to.

content/community/twitter.md
```diff
@@ -1,10 +1,10 @@
 +++
-title = "Twitter"
+title = "X"
 # https://en.wikipedia.org/wiki/Twitter
-description = "Twitter is a microblogging and social networking service on which users post and interact with messages known as \"tweets\""
+description = "X, is a microblogging and social networking service on which users post and interact with messages known as \"tweets\""
 date = "2022-07-23T03:51:55-04:00"
 draft = false
-logo = "https://upload.wikimedia.org/wikipedia/commons/4/4f/Twitter-logo.svg"
+logo = "https://upload.wikimedia.org/wikipedia/commons/c/ce/X_logo_2023.svg"
```
reclaimingmytime, line 7: This is using Wikipedia as a CDN, which might not be very reliable. The file name and path might get changed any time. For example, the Twitter logo has already been removed. Also, it might load slower than a local file.

CGBassPlayer: I agree with the idea of the local files wherever we can, the official twitter logo can be be downloaded [here](https://about.twitter.com/en/who-we-are/brand-toolkit)

Issue #154, opened 2022-07-22
Archive - populate with archived shows


gerbrent:
This is a big one, but essential for our 1.0 Milestone

FOr all shows listed on [jupiterbroadcasting.com's](https://www.jupiterbroadcasting.com/) Archive sub-menu:
* [ ] scrape show metadata: title, description, hosts, etc.
* [ ] create sub-menu items for each Show
* [ ] create Show page w all episodes listed (w pagination)
* [ ] implement Show host/guest pages - but NOT NECESSARILY also listed in all JB hosts/guests listings
* [ ]  ...?

### Potential Problems
Some shows....
* ...listed in the archive do not feature Show artwork, only per-episode artwork
   * [ ] design Show artwork - can be fairly simple & derivative of current episode art
* ...have a Fireside listing, some do not.
   * i.e. [Choose Linux](https://chooselinux.show/)
* ...are no longer hosted on the JB Network, do have archive items, but are still actively produced shows with the same Fireside site 
   * i.e. [BSD Now](https://www.bsdnow.tv/)
* ...feature outdated artwork branding 
   * i.e. [Choose Linux](https://chooselinux.show/), [User Error](https://error.show/) feature Linux Academy branding
   * [ ] modify branded artwork to modern JB design guidelines
* ...do not follow the current JB Episode naming conventions
   * [Computer Action Show](https://www.jupiterbroadcasting.com/show/cas/)

#### Filesystem
* Preferred files structure could be `/content/archive/show/beer-is-tasty` to keep the `/content/show` for active shows, i.e. cleaner for everyday use

#### Sub-menu?
I feel the sub-menu approach is a fine one, with all archived shows being presented under the "Archive" menu item. 
Open to discussion.

#### Media Files?
* [ ] audit media files and locations to inform...
* [ ] #23

#### Hosts/Guests
> implement Show host/guest pages - but NOT NECESSARILY also listed in all JB hosts/guests listings

We do not necessarily want all previous hosts and guests to be listed on the main Hosts and Guests pages - possibly only listing *active* show Hosts/Guests there. Worth discussion and decision/input further from JB.

..... more?

gerbrent:
Here's another helpful way Wes and I have been exploring the scope of this item:

* The archive *could* potentially be added after the 1.0 Milestone, but ultimate must be added within a reasonable time-frame.
[...]

PR #568, opened 2023-09-14
Feature/add cr episodes 0 54

### LINKED ISSUE #222: [Coder Radio] Episodes 1-54 Missing
I think this was mentioned in a show once regarding why these episodes aren't in Fireside (something about a migration from a different podcast hosting provider).

Looks like the audio files no longer exist on Libsyn (which looks like what was used prior) for these episodes, but the video does.

ChanceM:
Resolves #222 

Add content for episodes and add image (EP 38).
If podcast_file is None try to play direct files.


content/show/coder-radio/0000.md
```diff
@@ -0,0 +1,43 @@
...
+    "michael"
+  ],
+  "guests": [],
+  "sponsors": [],
+  "podcast_duration": null,
+  "podcast_file": null,
+  "podcast_bytes": null,
+  "podcast_chapters": null,
+  "podcast_alt_file": "https://traffic.libsyn.com/jnite/thedust-mp3.mp3",
+  "podcast_ogg_file": "https://traffic.libsyn.com/jnite/thedust-ogg.ogg",
```
reclaimingmytime, line 29: This link doesn't point to a file anymore, including anything with `libsyn.com/jnite` (Jupiter@Nite, the old JB show?)

content/show/coder-radio/0000.md
```diff
@@ -0,0 +1,43 @@
...
+  "video_file": "https://201206.jb-dl.cdn.scaleengine.net/coderradio/2012/thedust-432p.mp4",
```
reclaimingmytime, line 30: All `scaleengine.net` links from this PR are also not accessible.

reclaimingmytime:
To to make the old episodes accessible, we would have to re-host the old episodes on a different server. This is part of issue #154, along with all the other unmigrated JB shows.

ChanceM:
> To to make the old episodes accessible, we would have to re-host them on a different server. This is part of issue #154, along with all the other unmigrated JB shows.

Yes, my goal here was to mainly add back content to an active show and to keep it as close to the original content as possible. For that matter JB could rip the audio and add them all back to fireside.

content/show/coder-radio/0001.md
```diff
@@ -0,0 +1,71 @@
...
+  "jb_url": "/20497/gateways-to-programming-cr-01/index.html",
+  "fireside_url": "/1"
+}
+
+
+### Episode Links
+
+## Show Notes:
+
+#### What is Coder Radio?
```
CGBassPlayer, line 43: For Episodes that have both `Show Notes` and `Episode Links`, one should be removed so we don't have multiple titles for the same content.

CGBassPlayer (changes requested):
First of all. This is really impressive and is greatly appreciated. I am not sure how you got all the early coder episodes, but I love that you did! :slightly_smiling_face:  

I do think that the `Show Notes` title or the `Episode Links` title in the markdown needs be removed. Having both does not look that good.
image

The other issue I saw when testing the page is that the tags title still appears even when there is no tags. I already have code ready to fix this (just a quick conditional in the template)  so no need to worry about that.

Last thing @reclaimingmytime had already pointed out being the the links that are dead. Not sure where their new location is but we should probably ask JB if they still have them hosted somewhere.

CGBassPlayer (approved):
I think I can live with some of the links being dead for now. I think what we have now will be good enough for the time being and we can address the 404 links when the archive site starts.

PR #572, opened 2023-11-15
Restore Host and Guest Lists

### LINKED ISSUE #571: Hosts and Guests overview links to tags page
### URL

https://www.jupiterbroadcasting.com/hosts/ and https://www.jupiterbroadcasting.com/guests/

### Summary

_People --> Hosts_ and _People --> Guests_ links to a list of tags.
image

https://www.jupiterbroadcasting.com/people/ (clicking on "People") works.

CGBassPlayer:
Fixes #571 

The terms.html file seems to wipe out the host and guest pages. While I do not understand the inner workings of why that is the case, simply moving the html template to fall inline with the other list layouts seems to fix the issue while retaining the list of tags.


CGBassPlayer:
Converting to draft for now because the single tags now return the full list

CGBassPlayer:
So it seems like there was a file encoding issue of some kind because with no lines of code changed, my local instance is working as expected.

local instance vs jb.com

CGBassPlayer:
After playing with this way too late last night. I figured out that for some reason if I implements the fix with the dev server still running, It would appear to be fixed. If I restarted the dev server, the build would hang (that is why the tests keep failing as well). Going to be looking into this further.

CGBassPlayer:
So it was not file encoding. Turns out the location of the html file was a global taxonomy location. Moving the file to the tags folder means that the updated template now only affects the tags page.

CGBassPlayer:
I am going to go ahead and merge this so we can fix the issue with the tags being shown on the wrong pages

PR #590, opened 2024-03-22
🚨 New show needs added to jb.com

ChrisLAS:
I would like our new show, This Week in Bitcoin, to appear on the Jupiter Broadcasting website. 🎉

- Show website: https://www.thisweekinbitcoin.show/
- RSS: https://serve.podhome.fm/rss/55b53584-4219-4fb0-b916-075ce23f714e

The tricky thing here is that this is using a new platform Podhome.fm. So, I'm not sure what the challenges are in pulling it in vs a show published on Fireside. 

But I'd like to see this solved for This Week in Bitcoin because, given time, I could see us migrating some of our other shows off Fireside. Some of them might live on Podhome.FM.


CGBassPlayer:
Going to close this as I just synced main and develop. And I am not sure what Chris was doing in his map scramble. All related discussions can take place in #591

PR #618, opened 2024-05-17
Update prod deployment command

CGBassPlayer:
Updates the Makefile's prod deployment to use the newer `docker compose` command as well as add the `--force-recreate` flag so the container is recreated even if the container already exists.


ironicbadger:
LGTM

PR #620, opened 2024-05-18
Main

ironicbadger:


CGBassPlayer:
Just a heads up for the future: when after develop goes to main you can sync main into develop using `git pull origin main` while in the develop branch than just `git push`. Reduces the number of PRs we need.

PR #701, opened 2024-10-08
Add scraped episode(s)

ChanceM:


CGBassPlayer:
Unless I missed something, I am pretty sure the `summer` coupon code is no longer used so this PR probably shouldn't be merged.

ChanceM:
> Unless I missed something, I am pretty sure the `summer` coupon code is no longer used so this PR probably shouldn't be merged.

Yeah it's a race condition because the episode sponsors are parsed concurrently and not necessarily chronological.

PR #762, opened 2025-01-13
Add scraped episode(s)

ChanceM:

### DIFF
--- a/content/people/michaeltunnell.md
+++ b/content/people/michaeltunnell.md
@@ -1,13 +1,12 @@
 ---
 avatar: /images/people/michaeltunnell.jpg
-avatar_small: /images/people/michaeltunnell_small.jpg
-bio: Podcaster / Content Creator for the Destination Linux Podcast. I'm also a Designer,
-  Marketer, Writer, Open Sourcerer, Linux Enthusiast, and host of This Week in Linux.
+avatar_small: null
+bio: null
 homepage: https://tuxdigital.com/
 instagram: null
 linkedin: null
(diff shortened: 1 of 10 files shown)


content/people/michaeltunnell.md
```diff
@@ -1,13 +1,12 @@
 ---
 avatar: /images/people/michaeltunnell.jpg
-avatar_small: /images/people/michaeltunnell_small.jpg
-bio: Podcaster / Content Creator for the Destination Linux Podcast. I'm also a Designer,
-  Marketer, Writer, Open Sourcerer, Linux Enthusiast, and host of This Week in Linux.
+avatar_small: null
```
noblepayne, line 3: Are these needed/used by the website guest page? We can probably cleanup the entry some, just want to make sure we don't make the page look worse as a result.

Probs at least partially my fault, added a podcast:person entry for him in a recent ep, but it was quite minimal compared to some of our other info. Eventually we'll get this better unified.

noblepayne: and thank you for looking into this! :pray:

ChanceM: avatar_small is not currently used in the site and is limited to Fireside sites, and also not supported by [podcast:person](https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md#person). We could add something to the scraper to make a smaller thumbnail image but I don't know that it is needed.

CGBassPlayer (approved):

PR #777, opened 2025-02-11
Add Episodes to The Launch :rocket:

CGBassPlayer:
Includes:
- #775 
- Adds Episodes 5-8 for the new show
- Updates navbar test to look for new show

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -94,6 +94,13 @@ paginate = 12
         weight =97
         parent = "shows"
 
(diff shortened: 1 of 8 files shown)


CGBassPlayer:
Gonna get this into the dev version of the site. We just need a 16:9 image and to add this podcast to whatever podhome requires to allow us to have the show in the player. With it being in dev, we can easily check to see when it taken care of before pushing to production

PR #800, opened 2025-02-26
Add Picks to site

### LINKED ISSUE #763: A Pen For Picks, A Pick Pen.
### Feature Description

As far as I am aware there is no central repository or storage medium when it comes to the Picks from the Episodes. There is the Boost Barn for seeing all of the boosts for each episode of Self-Hosted and maybe LUP? I can't seem to find one for LUP but I must be overlooking it. 

The idea itself as I thought about (taking my programming abilities into account) was going to be a relatively simple list with each episode having its Pick(s) next to it. A short conversation with Wes in the Matrix chat, his take is 

_(a) a data set of some kind, probably with sufficient metadata to associate show and episode for the pick at least, or something like that and (b) a non-raw data display method of some kind (or multiple I guess), whether directly on the main site or elsewhere._

Which I like more, just not something I have done. Either way I wanted to get the idea out there to hopefully inspire others to join me. 

### Reference to Feature

_No response_

ChanceM:
Resolves #763

### DIFF
--- a/content/community/picks.md
+++ b/content/community/picks.md
@@ -1,0 +1,8 @@
++++
+title = "Picks "
+description = "Pick Pen"
+date = "2025-02-23T21:23:01-08:00"
+draft = false
+layout = "picks"
+type = "picks"
++++
--- a/data/picks/add-water.yaml
+++ b/data/picks/add-water.yaml
@@ -1,0 +1,13 @@
+---
+description: A utility app to easily install the Firefox GNOME Theme and automatically
+  update it in the background. This theme keeps Firefox fashionable within the GNOME
+  design ecosystem, and provides many helpful features to customize the interface
+  to make browsing even more pleasant
+license: GPL-3.0
+shows:
+- episode: 602
+  show: LINUX Unplugged
+  slug: linux-unplugged
+title: Add Water
+url: https://flathub.org/apps/dev.qwery.AddWater
+---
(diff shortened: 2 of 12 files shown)


CGBassPlayer:
Hey sorry for the delay! 

Would it make more sense to convert the picks files from YAML to markdown files and move the description to the content section? Just curious to see what you think and how much work it would be to do it that way. 

I'm just thinking just in case we want more formatting freedom in the future. What are your thoughts on feasibility/practicality on that?

ChanceM:
[Hugo Data Files](https://bwaycer.github.io/hugo_tutorial.hugo/extras/datafiles/#:~:text=Hugo%20supports%20loading%20data%20from,It%20even%20works%20with%20LiveReload.)

Hugo Data Files are a bit different from content. It supports YAML, JASON, or TOML. We can still access all elements and display them how ever we want. The benefit is that each item doesn't generate a page. We have thought and I have tested converting sponsors to a dataset as well which would reduce page generation for things we don't use.

CGBassPlayer:
Ah that makes sense! I forgot the content files generate individual pages no matter what so good call with the data files. 

The code itself looks good to me. Only thing I would add (and its not mandatory) is the new link to the dropdown test in `conftest.py`. It seems like the tests pass without it (which is interesting) but it should probably be there for the sake of consistency.

ChanceM:
Added the test parameter, but I also have a re-written test I will make a separate PR for that will actually check the dropdowns on the site against the expected values. This will ensure the test will fail for added items not added to the test parameters.

CGBassPlayer (approved):

PR #809, opened 2025-03-12
Fix test to compare page dowpdowns against expected dropdowns.

ChanceM:
This will check the sites dropdowns against the expected dropdowns. Previously we where only checking that the expected dropdown items where on the page.


ChanceM:
Just noting the test failed because the expected drop-down item is missing.

CGBassPlayer (approved):
After updating your branch this test is looking good.

PR #846, opened 2025-04-28
Add scraped episode(s)

ChanceM:


CGBassPlayer (approved):
Hoping these PR should stop coming in now that merged https://github.com/JupiterBroadcasting/show-scraper/pull/51

My working theory is that the old scraper and the new scraper are overwriting each other on the URL

PR #853, opened 2025-05-12
Add scraped episode(s)

ChanceM:

### DIFF
--- a/data/picks/element.fm.yaml
+++ b/data/picks/element.fm.yaml
@@ -1,0 +1,10 @@
+---
+description: Unlimited, GPLv3 Licensed, self-hostable podcast hosting
+license: null
+shows:
+- episode: 614
+  show: LINUX Unplugged
+  slug: linux-unplugged
+title: ELEMENT.FM
+url: https://element.fm/
+---
--- a/data/picks/meshsidecar.yaml
+++ b/data/picks/meshsidecar.yaml
@@ -1,0 +1,12 @@
+---
+description: meshSidecar provides a flexible way to integrate mesh networking capabilities
+  with NixOS services. It allows services to participate in mesh networks without
+  requiring direct modification of the service itself.
+license: null
+shows:
+- episode: 614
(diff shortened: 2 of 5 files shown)


CGBassPlayer (approved):

data/picks/element.fm.yaml
```diff
@@ -0,0 +1,10 @@
+---
+description: Unlimited, GPLv3 Licensed, self-hostable podcast hosting
+license: null
```
CGBassPlayer, line 3: Automation can't pick up
That this is GPLv3 so we should try to remember that later

data/picks/meshsidecar.yaml
```diff
@@ -0,0 +1,12 @@
+---
+description: meshSidecar provides a flexible way to integrate mesh networking capabilities
+  with NixOS services. It allows services to participate in mesh networks without
+  requiring direct modification of the service itself.
```
CGBassPlayer, line 4: I don't think the whole description will be posted on the site. This might need to change to a single line.

data/picks/element.fm.yaml
ChanceM: It was never programmed to do so, as there is no structured way to do it using the current methods. My hope was in the future we would use a JB namespace in the custom XML. But I tried to make the dataset as an example so if some were put in manually they would have a place to populate that data.

data/picks/meshsidecar.yaml
ChanceM: Yes it will display.

data/picks/element.fm.yaml
CGBassPlayer: I figured there isn't, just wanted to note that we should update this file manually at some point

noblepayne: Thank you both for the attention to detail! Def agreed on moving to a proper encoding of picks in the feed with license etc.

PR #875, opened 2025-06-24
Template and link updates for TWiB on Fountain

noblepayne:
With TWiB moving to Fountain we'd like to use the JB.com show page as the _main_ page for the show. To support this Chris requested some link cleanup:
- Remove support link
- Remove contact link
- Remove shownotes link
- Add episodes.fm link
- Add direct RSS feed link.

To try and keep this minimally invasive functionality wise I moved the support link into the optional links and added it to all non-twib shows. Realistically, we may be fine just dropping it universally instead. In this version all that changes for non-twib shows is the ordering of the links/icons in that the support one is now further to the right. 

Open to any changes, this was just meant as a start.

image

### DIFF
--- a/content/show/this-week-in-bitcoin/_index.md
+++ b/content/show/this-week-in-bitcoin/_index.md
@@ -14,10 +14,10 @@ header_image = "/images/shows/this-week-in-bitcoin.png"
 
 podverse_podcast_id = "MxEmZU_zDt"
 
-[links.shownotes]
-  url="https://www.thisweekinbitcoin.show"
-[links.email]
-  url="/contact"
+[links.episodesfm]
+  url="https://episodes.fm/1733733790"
+[links.feed]
+  url="https://feeds.jupiterbroadcasting.com/twib"
 
 +++
 
--- a/content/show/this-week-in-bitcoin/_index.md
+++ b/content/show/this-week-in-bitcoin/_index.md
@@ -14,7 +14,7 @@ header_image = "/images/shows/this-week-in-bitcoin.png"
 
 podverse_podcast_id = "MxEmZU_zDt"
 
-[links.episodesfm]
+[links.smart-link]
   url="https://episodes.fm/1733733790"
 [links.feed]
   url="https://feeds.jupiterbroadcasting.com/twib"
--- a/themes/jb/layouts/partials/show/links.html
+++ b/themes/jb/layouts/partials/show/links.html
@@ -3,14 +3,13 @@
   <a class="link level-item pr-3" target="_blank" href="{{ (.GetPage `subscribe.md`).Permalink }}" title="subscribe">
     <div class="title fa fa-bell-ringing-o fa-3x"><span class="sr-only">subscribe</span></div>
   </a>
-  <a class="link level-item pr-3" target="_blank" href="{{ (.GetPage `/membership`).Permalink }}" title="support">
-    <div class="title fa fa-usd fa-3x"><span class="sr-only">support</span></div>
-  </a>
   {{with .Params.links}}
     {{ range $key, $value := . }}
       <a class="link level-item pr-3" target="_blank" href="{{$value.url}}" title="{{$key}}">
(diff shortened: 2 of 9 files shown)


content/show/this-week-in-bitcoin/_index.md
```diff
@@ -14,10 +14,10 @@ header_image = "/images/shows/this-week-in-bitcoin.png"
 
 podverse_podcast_id = "MxEmZU_zDt"
 
-[links.shownotes]
-  url="https://www.thisweekinbitcoin.show"
-[links.email]
-  url="/contact"
+[links.episodesfm]
+  url="https://episodes.fm/1733733790"
```
CGBassPlayer, line 18: I wonder if this should be more of the individual show site link since it uses the `podcast` icon. I think that would look better than the `shownotes` link that we use currently.

```suggestion
[links.show-site]
  url="https://episodes.fm/1733733790"
```

themes/jb/layouts/partials/show/links.html
```diff
@@ -21,6 +20,10 @@
           <div class="title fa fa-{{$key}} fa-3x"><span class="sr-only">{{$key}}</span></div>
         {{ else if eq $key "wiki" }}
           <div class="title fa fa-book fa-3x"><span class="sr-only">{{$key}}</span></div>
+        {{ else if eq $key "episodesfm" }}
+          <div class="title fa fa-podcast fa-3x"><span class="sr-only">{{$key}}</span></div>
```
CGBassPlayer, line 24:
```suggestion
        {{ else if eq $key "show-site" }}
          <div class="title fa fa-podcast fa-3x"><span class="sr-only">{{$key}}</span></div>
```

CGBassPlayer (commented):
Just some thoughts

content/show/this-week-in-bitcoin/_index.md
CGBassPlayer: I just feel like episode.fm might be too specific (much like how we have fireside_url in episode pages even though most of our shows don't come from there anymore). I just want it to stay generic enough to where we don't need to update it to something else down the road if we don't have to.

noblepayne: > I just want it to stay generic enough to where we don't need to update it to something else down the road if we don't have to.

Totally on board with this. I wonder about names... Thinking about what episodesfm is doing maybe some ideas:
- listen link
- podcatcher
- app link
- directory
- universal link
- smart link

But I'm fine with whatever.

noblepayne:
While we are at it, or in a follow-up PR, I think we probably want to add episodes.fm links to all the shows (or at least all current shows).

content/show/this-week-in-bitcoin/_index.md
noblepayne: @CGBassPlayer thoughts on any of these? Still prefer `show-site`? I'm down for whatever just want to poke this along.

CGBassPlayer: I like the `universal-link` or the `smart-link`

I think either of those would work well and that would be good if you guys move away from episodes.fm for whatever reason

CGBassPlayer: Just updated my suggestions to `smart-link`

CGBassPlayer:
> add episodes.fm links to all the shows (or at least all current shows).

Do we have all those links ready to go?

content/show/this-week-in-bitcoin/_index.md
noblepayne: Updated to smart-link.

themes/jb/layouts/partials/show/links.html
noblepayne: Updated to `smart-link`

noblepayne:
> Do we have all those links ready to go?

We did not previously, but I just grabbed them and added them here.

PR #882, opened 2025-06-29
Add scraped episode(s)

ChanceM:


CGBassPlayer:
With the new twib feed I can see we are gonna lose the value splits and some links. Not sure if we want to fix that before this is merged?

noblepayne:
>  but at least the value block would be nice to keep

Unfortunately at present at least the valueTimeSplits are gone from the feed. Though maybe we would prefer not to delete them from our copy? Pros and cons to that... 

Also looks like podhome put a lot of stuff at item level even if not strictly needed (e.g. it seemed to have a per-show image even if it was just a copy of the default image). So could be that some split "deletions" are versions like that, where fountain just leaves it at the show level unless actually overridden per item.

ChanceM:
We can exclude the past episodes from overwrite in the scraper.

PR #892, opened 2025-07-24
Replace Selfhosted episodes.fm with the launch

ChanceM:
The Launch had the link for Selfhosted. Correcting for The Launch.


noblepayne:
Great catch, thank you!

PR #911, opened 2025-09-01
Replace live page, add audio/video toggle

noblepayne:
This commit replaces the old live page with a new implementation that includes a toggle between audio and video streams.

Key changes:
- Introduces an audio/video player toggle on the /live page.
- Adds stream URLs to the site config.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -7,6 +7,10 @@ theme = 'jb'
 
 paginate = 12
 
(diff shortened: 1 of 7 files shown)


noblepayne:
Here are screenshots of the new version in use:

Audio mode:
livepage_audio

Video mode:
livepage_video

PR #1021, opened 2026-03-31
Add missing licenses (and one description) for picks

eroc1990:
This PR adds licenses to multiple picks that were previously missing from the `license:` field.


CGBassPlayer (approved):
LGTM, Thanks for adding those!

PR #1036, opened 2026-04-22
Move VPS deployment logic to standalone script

noblepayne:
Currently, the deployment logic is embedded directly in the YAML workflow, which makes it hard to maintain, test, or lint. It also had a few issues where a failed git pull would silently continue the build.

This moves the deployment logic to `scripts/deploy.sh` and pipes it to the VPS.

Notable changes:
- Switched `git pull` to `git fetch` + `reset --hard` to avoid merge conflicts on the remote.
- Added `set -euo pipefail` so the pipeline actually stops if a command fails.
- Moved `GH_TOKEN` from a command argument to an environment variable to keep it out of the process list.
- Added a 15-minute job timeout so it doesn't hang indefinitely if the connection drops.
- Added `echo` markers to the script to show progress in the Action logs.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -20,8 +20,9 @@ on:
 jobs:
   deploy-via-tailscale:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
-      # deploys using alexktz tailnet
+      # deploys using JB GitHub tailnet
       - name: Connect Tailscale
         uses: tailscale/github-action@v2
         with:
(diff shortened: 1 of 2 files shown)


CGBassPlayer (approved):
LGTM. I didn't realize we pushed the built images to the repo. They are private by default so no one outside the JB org members could see it. Not a big deal but not sure if anyone was aware

noblepayne:
> LGTM. I didn't realize we pushed the built images to the repo. They are private by default so no one outside the JB org members could see it. Not a big deal but not sure if anyone was aware

Thanks for looking! And yeah, had a similar reaction. Probably not needed, but not hurting anything for now. :man_shrugging:  I'd love to (at some point...) take a wider (and perhaps nixier) pass on some of this deploy stuff, so perhaps re-address then.

